Issue #1, opened 2017-11-16
Setup Skeleton


mateoclarke:
- [ ] ~ES Lint `.eslintrc`~
- [x] `.editorconfig`
- [x] `package.json` has description and version
- [x] add `Dockerfile`
- [x] add `.dockerignore`
- [x] add `gitignore`
- [ ] ~pick `license`~

optionals:
- `CONTRIBUTING.md`
- `ISSUE_TEMPLATE.md`

mateoclarke:
Let's punt on eslint since create react app comes with some basic linting config in the command line and browser console.

Let's punt in license adoption until we have a discussion about it.

PR #9, opened 2017-11-20
Docker setup

mateoclarke:
completes #1


docker-compose.yml
```diff
@@ -0,0 +1,7 @@
+app:
+  volumes:
+    - "./:/app"
+    - "/app/node_modules/"
```
mateoclarke: One problem that @ifsimicoded ran into when she tried docker setup on her local machine was it wouldn't work unless she had prerun `yarn install` before any of the `docker-compose` commands. She discovered this trick on line 4 that apparently prevents the node_modules directory in the docker container from getting overwritten or blown away when the volume referencing happens. 🤷‍♂️

README.md
```diff
@@ -284,7 +304,7 @@ In the WebStorm menu `Run` select `Edit Configurations...`. Then click `+` and s
 
 Start your app by running `npm start`, then press `^D` on macOS or `F9` on Windows and Linux or click the green debug icon to start debugging in WebStorm.
 
-The same way you can debug your application in IntelliJ IDEA Ultimate, PhpStorm, PyCharm Pro, and RubyMine. 
+The same way you can debug your application in IntelliJ IDEA Ultimate, PhpStorm, PyCharm Pro, and RubyMine.
```
mateoclarke: blah, trailing spaces shit. _sorry_ This reminds me we need an editorconfig

mateoclarke: done ✨

docker-compose.yml
mateoclarke: The downside to this we've now discovered is that whenever you add new dependencies in `package.json` you have to rebuild your image. To do this, we are running `docker-compose build --no-cache`. Is there a better way to set this up in docker?

ifsimicoded:
Hey took some time to screw around with this and came up with the setup in this branch (built off this branch -- docker_setup). It's simpler for getting node_modules imported without having to rebuild the docker image.

https://github.com/cityofaustin/janis/tree/sd_docker_setup

docker-compose.yml
benweatherman: > There seems to be an important order of operations on copy over files into the volume and running yarn install

The `Dockerfile` gets compiled (which [runs `yarn`](https://github.com/cityofaustin/janis/pull/9/files#diff-3254677a7917c6c01f55212f86c57fbfR5)). So the compiled docker image is good. Then running `docker-compose up` or `docker-compose run` will mount the volumes. In this case you're mounting the whole local directory `.` to the `/app` directory inside the container. So unless you've run `yarn` locally you won't have any `node_modules`.

> She discovered this trick

🎉 nice work @ifsimicoded! This will create a persistent volume that ensures `node_modules` doesn't get mounted from the local filesystem. This seems like an elegant solution.

> Is there a better way to set this up in docker?

Running `docker-compose build --no-cache` seems like the right thing if `docker-compose up --build` doesn't work correctly. I'd say don't use `docker-compose` until you need to. I'd also suggest a `serve-local.sh` or similar command that people can just run without needing to remember the secret incantation from project to project.

README.md
```diff
@@ -1,3 +1,23 @@
+# janis
+Frontend for [joplin](https://github.com/cityofaustin/joplin).
+
+## Run locally
+
+```
+docker-compose build
```
benweatherman: How do you feel about adding a `scripts/serve-local.sh` file that can contain whatever commands need to be run? If we use that consistently across projects it'll be an easy way to start working on a new project.

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+indent_size = 4
+
+[*.{js,json,yml}]
+indent_size = 2
```
benweatherman: nit: since this is a JS project, should we make the default indent 2?

Dockerfile
```diff
@@ -0,0 +1,7 @@
+from node
+RUN mkdir /app
+WORKDIR /app
+COPY package.json /app/package.json
```
benweatherman: We also need to copy `public` and `src`. The `Dockerfile` needs to contain everything we need to run the server without mounting local volumes/directories.

(sorry this wasn't part of my review)

README.md
mateoclarke: The only question I have about this from looking at the `serve-local.sh` script you have in joplin, it seems like you would have to run docker build each time you start the local server. Is that a problem at all?

Dockerfile
```diff
@@ -0,0 +1,7 @@
+from node
+RUN mkdir /app
+WORKDIR /app
+COPY [ "package.json", "src", "public", "/app/"]
+RUN yarn
+EXPOSE  3000
```
benweatherman: Not sure if this is a good idea or not: if we always expose port 80 for web servers, it'll make deployment slightly easier across projects. It does make things slightly more confusing as mentioned in cityofaustin/joplin#7

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+indent_size = 2
+
+[*.{js,json,yml}]
```
benweatherman: nit: I'd say nix this part since it's the same `indent_size`

Dockerfile
mateoclarke: Sorry, I don't follow what you're suggesting here. Should line 6 be `EXPOSE 80` instead?

.editorconfig
mateoclarke: cool, since I already merged I just updated that on master

Dockerfile
benweatherman: Yes that's what i'm suggesting

PR #19, opened 2017-11-21
Docker setup fix for node_modules

ifsimicoded:
update docker setup script to mount specific dirs into container to avoid node_modules dir being overwritten


scripts/serve-local.sh
```diff
@@ -10,5 +10,6 @@ docker run \
     --name janis \
     --tty --interactive \
     --publish 3000:3000 \
-    --volume "$PWD:/app" \
+    --volume "$PWD/src:/app/src" \
+    --volume "$PWD/public:/app/public" \
```
benweatherman: Does it make sense to put `package.json` in there as well? Most likely you'd need to rebuild if you add a package, but there might be other changes that wouldn't require a rebuild.

While we're talking about it, I'm also wondering if we need to copy `yarn.lock` in `Dockerfile`?

ifsimicoded: I'll add these.

package.json
```diff
@@ -10,7 +10,7 @@
     "react-scripts": "1.0.17"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "PORT=80 react-scripts start",
```
ifsimicoded: @benweatherman not sure if this is the best place to tell react not to use the default 3000 port. Or if it's better to set the PORT. Nor am I even sure if this is available to change for production.
https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#advanced-configuration 

Something we can screw around with later, unless you have experience here.

benweatherman: A `.env` file might be our best bet so we don't need to configure it for each environment. We'd have to make sure we're copying that file into our docker image. It might also be more trouble than it's worth to use the non-default port so maybe 3000 is fine.

Chef's choice 😸

scripts/serve-local.sh
```diff
@@ -9,6 +9,10 @@ docker run \
     --rm \
     --name janis \
     --tty --interactive \
-    --publish 3000:3000 \
-    --volume "$PWD:/app" \
+    --publish 3000:80 \
+    --volume "$PWD/src:/app/src" \
+    --volume "$PWD/public:/app/public" \
+    --volume "$PWD/package.json:/app/package.json" \
+    --env PORT=80 \
```
ifsimicoded: @benweatherman please don't hate me for the back and forth in this PR. I blame the current administration. 
This seems like the cleanest route to set environment vars as .env files aren't acknowledged universally (ran into NODE_PATH issues with storybook on ctxfloods) and it doesn't litter up our package.json. Let me know if you think otherwise. I'm definitely out of my party zone.

Dockerfile
```diff
@@ -1,7 +1,7 @@
 from node
 RUN mkdir /app
 WORKDIR /app
-COPY [ "package.json", "src", "public", "/app/"]
+COPY [ "yarn.lock", "package.json", "src", "public", "/app/"]
 RUN yarn
-EXPOSE  3000
+EXPOSE 80
```
mateoclarke: gracias 😸

scripts/serve-local.sh
benweatherman: This seems like a great way to go too. It'll either work or we'll change it 😸

scripts/serve-local.sh
```diff
@@ -1,14 +1,24 @@
 #!/usr/bin/env bash
+echo "starting..."
 
 set -o errexit
 
-TAG='janis:local'
+echo "fetching node packages on host..."
+mkdir -p .yarn-cache
+yarn install --cache-folder .yarn-cache
```
benweatherman: We run the risk of building something for the host machine that isn't compatible with the docker OS. I'm starting to think the benefits of the cache on the host isn't that beneficial compared to the work involved to make it happen.

Dockerfile
```diff
@@ -1,7 +1,13 @@
...
-RUN yarn
-EXPOSE  3000
+ENV PORT=80
+ENV NODE_PATH=src
+COPY yarn.lock /app/yarn.lock
+COPY .yarn-cache /.yarn-cache
+COPY package.json /app/package.json
+COPY public /app/public
+COPY src /app/src
+RUN yarn install --cache-folder /.yarn-cache
```
benweatherman: This should go immediately after copying `yarn.lock` & `package.json`. That way we'll only reinstall when those files change. Currently the dependencies will be reinstalled if any of the `src` or `public` files change 🙀

.dockerignore
```diff
@@ -0,0 +1,6 @@
+*
+!public/
+!src/
+!package.json
+!yarn.lock
+!.yarn-cache
```
benweatherman: I think we're adding newlines at the end of files. Install the `editorconfig` package for sublime text and it'll add them automatically.

scripts/serve-local.sh
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
+echo "starting..."
 
 set -o errexit
 
 TAG='janis:local'
-
+echo "building docker image..."
```
ifsimicoded: I originally added this echo for debugging and just left it as I found after multiple docker stops and starts, docker build would hang for lengthy periods even after container/image/layer cache clean up. So this will just let a dev know something is happening.

PR #23, opened 2017-12-04
Services and Service Mobile Views

ifsimicoded:
first stage of JSX, CSS for service and services page.
We can breakdown css and jsx into smaller components as we go. 
This is just simplest for MVP development with just a single dev.
I also left the contact info on every page which can be easily removed by removing it from the json -- just seemed like useful info to have.


src/App.js
```diff
@@ -19,7 +18,7 @@ class App extends Component {
         <div>
           <Banner />
           <Header />
-          <section className="usa-section usa-grid">
+          <section>
```
mateoclarke: If we don't want to use `usa-grid` maybe we can start moving towards http://flexboxgrid.com?

src/__tmpdata/service_2.json
```diff
@@ -0,0 +1,46 @@
...
+    "Tuesday":"8:00 AM - 4:30 PM",
+    "Wednesday":"8:00 AM - 4:30 PM",
+    "Thursday":"8:00 AM - 4:30 PM",
+    "Friday":"8:00 AM - 4:30 PM",
+    "Saturday": "closed",
+    "Sunday":"closed"
+  },
+  "app": {
+    "title": "Recycle and Reuse Drop-off Center",
+    "type": "map",
```
mateoclarke: 👍

src/components/FormFeedback.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from 'react';
+
+class FormFeedback extends Component {
```
mateoclarke: Out of the scope of this PR, but maybe we think about what a more agnostic Form component would look like. We could build it in a way that is know how to consume a standard json schema. ex: https://github.com/mozilla-services/react-jsonschema-form

src/pages/Service.js
```diff
@@ -0,0 +1,145 @@
...
+    const related = get(data, "related", null);
+
+    return (
+      <div>
+
+        <div className="coa-section">
+          <a className="coa-page_breadcrumb" href={`/services/topic/${topicId}`}>{topicName}</a>
+          <h3 className="coa-page_title">{title}</h3>
+          { steps && (
+              <ol>{ steps.map((step) => { return <li>{step}</li> }) }</ol>
```
mateoclarke: this doesn't matter at all, but I like...
`<ol>{ steps.map((step) => <li>{step}</li>) }</ol>`

src/pages/Service.js
```diff
@@ -0,0 +1,145 @@
...
+
+      { body && (
+        <div className="coa-section" dangerouslySetInnerHTML={{__html: body}} />
+      )}
+
+      { (phone || email || address || hours) && (
+        <div className="coa-section">
+          <h4>Contact</h4>
+
+        { phone && (
```
mateoclarke: Each of these little displays could be their own component in the future.

src/pages/Service.js
```diff
@@ -0,0 +1,145 @@
...
+            <table className="usa-table-borderless">
+              <thead className="usa-sr-only">
+                <tr>
+                  <th scope="col">Day</th>
+                  <th scope="col">Open - Close Hours</th>
+                </tr>
+              </thead>
+              <tbody>
+              {
+                Object.entries(hours).map((hour, index) => {
```
mateoclarke: cool. wonder how `.entries` works differently than https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

src/pages/Service.js
```diff
@@ -0,0 +1,145 @@
...
+        )}
+
+        </div>
+      )}
+
+      { related && (
+        <div className="coa-section coa-section--grey">
+          <h4 className="coa-section__title">Use related services</h4>
+          {
+            related.map((service) => {
```
mateoclarke: These `coa-list__links` could be their own component

src/pages/Services.js
```diff
@@ -0,0 +1,88 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      title: '',
+      body: ''
+    };
+  }
+
+  componentDidMount() {
+    axios
```
mateoclarke: as we find ourselves writing more axios calls, we could move whatever is common into a utility module. Something like `getDataFromJoplin(pageId)`. Then we can include any functions we need in our components from that utility module and push more logic out of the component.

src/pages/Services.js
```diff
@@ -0,0 +1,88 @@
...
+      .get('http://localhost:8000/api/v2/pages/?format=json&type=base.ServicePage&fields=content,extra_content,tags')
```
mateoclarke: env var

src/pages/Services.js
```diff
@@ -0,0 +1,88 @@
...
+                  <i className="fa fa-chevron-right" aria-hidden="true"></i>
+                </a>
+              );
+            })
+          }
+        </div>
+
+        <div className="coa-section coa-section--grey">
+          <FormFeedback />
+          <a className="coa-section__link" href="#">Return to Top</a>
```
mateoclarke: This could be its own component. I might even cheat and check the ecosystem:
https://github.com/fisshy/react-scroll

src/pages/Services.js
```diff
@@ -0,0 +1,88 @@
...
+        </div>
+
+        <div className="coa-section">
+          <h4 className="coa-section__title">Request 311 Service or Call <a className="nowrap" href="tel:512-974-2000">512-974-2000</a></h4>
+          <p className="usa-content">311 is the city of Austin’s 24 hour information desk.</p>
+
+        {
+          services311.map((service) => {
```
mateoclarke: ditto: own component

PR #39, opened 2017-12-14
Sd component formvalidation

ifsimicoded:
- form validation
- clean up json stub data
- remove components in development (header and apps) for user testing


src/components/ContentItems.js
```diff
@@ -10,13 +10,13 @@ class ContentItems extends Component {
...
         <div className="coa-content_items" key={content.id}>
-          {(content.type === 'application_block') && (
+          {/*(content.type === 'application_block') && (
             <div className="coa-section">
               <h4>{content.value.description}</h4>
               INSERT {content.type} app HERE
               { content.value.url && <div dangerouslySetInnerHTML={{__html: content.value.url}} /> }
             </div>
-          )}
+          )*/}
```
mateoclarke: ewww. Is it ok if we delete this code instead of commenting it out? Or was this just a quick fix for usability testing? Maybe we add a `{/* TODO */}` or something easy to Command-F?

src/components/FormFeedback.js
```diff
@@ -1,35 +1,125 @@
...
       stage: 1,
+      values: {},
+      errors: {}
     };
   }
 
-  setStage = (stage) => {
+  handleFieldChange = (e) => {
+
+    const values = clone(this.state.values);
```
mateoclarke: This is the es6 way to do this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

src/components/FormFeedback.js
```diff
@@ -1,35 +1,125 @@
...
+    if (!value || /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/.test(value)) return false;
+    return 'Please enter a valid email.';
+  },
+  'required': (value) => {
+    if (typeof value !== 'undefined' && value !== null && /^\S/.test(value)) return false;
+    return 'This field is required.';
+  }
+};
+
+const formConfig = {
```
mateoclarke: I like it!

src/components/FormFeedback.js
mateoclarke: I do like the simplicity of the lodash syntax. Either way you do it, copy/clone is the way to go. 👏 
https://stackoverflow.com/questions/29537299/react-how-do-i-update-state-item1-on-setstate-with-jsfiddle#38378350

src/css/App.scss
```diff
@@ -48,6 +48,19 @@ ol, ul {
   }
 }
 
+.usa-input-error {
```
mateoclarke: Why do we use `usa-` namespace here? If it's something we write, shouldn't we use `coa-`?

src/css/_coa_variables.scss
```diff
@@ -6,8 +6,10 @@ $coa-space-level1: 5px;
 $coa-space-level2: 10px;
 $coa-space-level3: 15px;
 $coa-space-level4: 20px;
-$coa-space-level5: 40px;
-$coa-space-level6: 60px;
+$coa-space-level5: 25px;
+$coa-space-level6: 30px;
+$coa-space-level7: 40px;
+$coa-space-level8: 60px;
```
mateoclarke: How are we deciding these spacings? Should they be px? or rem?
Systems I've liked in the past do something like:
```css
coa-spacing: 10px;
coa-spacing-half: 5px;
coa-spacing-times2: 20px;
coa-spacing-times3-half: 35px;
```
that way the number units stay in sync. 🤷‍♂️

src/css/App.scss
```diff
@@ -288,6 +305,7 @@ label {
 }
 
 .coa-page_content {
+  .coa-overlay__content,
```
mateoclarke: This file is getting way to huge. Can we start to break it up into smaller .scss files plz? 🙏

mateoclarke:
At first, I requested changes, but I wouldn't hold off on merging. Just want to make sure we start a discussion about those points.

src/components/FormFeedback.js
ifsimicoded: I tend to like lodash for it's clarity -- i know what clone means, but Object.assign is a little esoteric. But eh, one less include, so I'll update

src/components/ContentItems.js
ifsimicoded: Yea, this was removed for usertesting. I can update with a comment.

src/css/App.scss
ifsimicoded: YASSSSS, I was reticent to pick a structure without your input. I'll take a stab at it in my next PR so this one doesn't get unruly.

src/css/_coa_variables.scss
ifsimicoded: I NEEEEEEED to convert these to rems for sure. I picked this naming convention to keep the name detached from the value, so if we change our minds in the future, I just have to update a rem value, and not all the variable names and their references. Also, I'm pulling these values from spacings in Celine's mocks. 

Honestly, I think space-xs, space-s, space-m, space-l, space-xl is the real way to go. This will also force us to limit the variety of spacings to only 5 which will keep things visually consistent. I just haven't thought much of it since it seems like the mocks are still in flux.

ifsimicoded:
@benweatherman, updated the validation to only validate the field in context on change instead of all the fields in the current stage. It's still done onChange, but can easily be changed to onBlur. But I figure this is something for user tests to decide.

ifsimicoded:
giphy-downsized

src/css/_coa_variables.scss
mateoclarke: yeah, tshirt size FTW! lets try that!

src/components/FormFeedback.js
mateoclarke: yeah, I go back and forth. My impulse is to use what JS gives you out of the box. But I also prefer clean reading and lodash/underscore are so well documented and widely used that they might as well be core JS. Really meant to just share with you that its now possible to do in ES6, but I'd leave it to chef's choice on which function to use in the future.

src/css/App.scss
ifsimicoded: I JUST remembered i never responded to this. we're overwriting styles for a uswds component here, it's not our own component.

mateoclarke: cool, can we put these overrides maybe in the _uswds_variables file? Or another good place?

PR #40, opened 2017-12-15
Footer

mateoclarke:
screen shot 2017-12-15 at 2 29 22 pm
screen shot 2017-12-15 at 2 29 36 pm

For SVGs, I went this approach: https://css-tricks.com/creating-svg-icon-system-react/

It takes a little more time to setup than just pulling in a npm module or using an font icon library, but will give us a lot of flexibility to decide what props we might want to give each SVG to customize them. It also seems like SVG with the proper aria labeling is a win for a11y.

Also, on a11y, noting in the code the aria labeling for links that pop new windows.
https://accessibility.18f.gov/


src/components/Footer.js
```diff
@@ -0,0 +1,45 @@
...
+  render() {
+    return (
+      <footer className="coa-Footer">
+        <div className="coa-section row">
+          <div className="col-xs-12">
+            <h3>AUSTIN.GOV</h3>
+          </div>
+          <div className="col-xs-12 col-md-6 last-md">
+            <p className="coa-Footer__body-text">
+              Austin.gov is a new website and a work in progress. For the full City of Austin website, visit <a href="http://austintexas.gov" target="_blank" aria-label="Opens in new window">austintexas.gov <ExternalLinkSVG size="13"/></a>. Learn more about the new website at <a href="http://projects.austintexas.io/projects/austin-digital-services-discovery/about/what-we-are-doing/" target="_blank" aria-label="Opens in new window">projects.austintexas.io <ExternalLinkSVG size="13"/></a>.
```
mateoclarke: when we use `target="_blank"` on an anchor tag, we should add `aria-label="Opens in new window"`. Wonder if there is a programmatic way to do that, or at least lint for it.
https://accessibility.18f.gov/

src/components/Footer.js
```diff
@@ -0,0 +1,45 @@
...
+import GithubSVG from '../svg/Github';
+import ExternalLinkSVG from '../svg/ExternalLink';
+import '../css/components/Footer.css';
+
+class Footer extends Component {
+
+  render() {
+    return (
+      <footer className="coa-Footer">
+        <div className="coa-section row">
```
ifsimicoded: since coa-section has margins on it, we can't use row in the same element, since row has conflicting margins. I know it'll result in a lot of nesting, so I'm open to suggestions as to how to handle this better.

src/components/Footer.js
mateoclarke: sure thing!

src/components/Footer.js
mateoclarke: hmmm, nesting isn't the worst at this point so I'll go with that.

src/css/components/Footer.scss
```diff
@@ -0,0 +1,30 @@
+@import '../coa_variables';
```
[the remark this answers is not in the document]
mateoclarke: I feel pretty ok about that. I hesitated to do it the way I did since you hadn't created a components scss folder yet. I was thinking about floods pain too.

ifsimicoded:
200w_d

mateoclarke:
For the comment: 
> Can you throw a .wrapper div around the footer so it maxes out at 1200px.

I think we want the grey background to stretch across but I agree we could max out the content. More nested divs i guess. 😬

mateoclarke:
That 🍌 gif grosses me out. Well done.

PR #42, opened 2017-12-18
Use location & contact from API

benweatherman:
Also changed API key from `theme` -> `topic`

### DIFF
--- a/src/components/Hours.js
+++ b/src/components/Hours.js
@@ -19,10 +19,10 @@ class Hours extends Component {
           </thead>
           <tbody>
           {
-            Object.entries(hours).map((hour, index) =>
+            hours.map((hour, index) =>
               <tr key={index}>
-                <th scope="row">{hour[0]}</th>
-                <td>{hour[1]}</td>
+                <th scope="row">{hour.day_of_week}</th>
+                <td>{hour.start_time} - {hour.end_time}</td>
               </tr>
             )
           }
--- a/src/pages/Service.js
+++ b/src/pages/Service.js
@@ -21,7 +21,7 @@ class Service extends Component {
 
   componentDidMount() {
     axios
-      .get(`${process.env.REACT_APP_CMS_ENDPOINT}/pages/${this.props.match.params.id}?fields=content,extra_content,theme(text)`)
+      .get(`${process.env.REACT_APP_CMS_ENDPOINT}/pages/${this.props.match.params.id}?fields=content,extra_content,topic(text),locations(location(name,street,city,state,zip,country,hours)),contacts(contact(name,email,phone))`)
       .then(res => {
         this.setState({ data: res.data })
       })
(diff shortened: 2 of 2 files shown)


src/pages/Service.js
```diff
@@ -21,7 +21,7 @@ class Service extends Component {
 
   componentDidMount() {
     axios
-      .get(`${process.env.REACT_APP_CMS_ENDPOINT}/pages/${this.props.match.params.id}?fields=content,extra_content,theme(text)`)
+      .get(`${process.env.REACT_APP_CMS_ENDPOINT}/pages/${this.props.match.params.id}?fields=content,extra_content,topic(text),locations(location(name,street,city,state,zip,country,hours)),contacts(contact(name,email,phone))`)
```
ifsimicoded: Live preview currently seems very coupled to the service page. Any plans for how we might achieve preview for all pages?
Also, would postMessage work for passing content to the iframe in a cleaner/safer way?
https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

src/pages/Service.js
```diff
@@ -31,15 +31,15 @@ class Service extends Component {
...
+    const topicId = get(data, "topic.id", null);
+    const topicName = get(data, "topic.text", null);
     const title = get(data, "title", null);
     const steps = get(data, "content", null);
     const contentItems = get(data, "extra_content", null);
-    const phone = get(jsonFileData, "contact.phone", null);
-    const email = get(jsonFileData, "contact.email", null);
-    const address = get(jsonFileData, "contact.address", null);
-    const hours = get(jsonFileData, "contact.hours", null);
+    const phone = get(data, "contacts[0].contact.phone", null);
```
ifsimicoded: Instead of picking off the just the first contact item, should we loop through the list and render multiple Contact components?

src/components/Hours.js
```diff
@@ -19,10 +19,10 @@ class Hours extends Component {
           </thead>
           <tbody>
           {
-            Object.entries(hours).map((hour, index) =>
+            hours.map((hour, index) =>
               <tr key={index}>
-                <th scope="row">{hour[0]}</th>
-                <td>{hour[1]}</td>
+                <th scope="row">{hour.day_of_week}</th>
+                <td>{hour.start_time} - {hour.end_time}</td>
```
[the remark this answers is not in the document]
benweatherman: CITY SERVICES DON'T GET TO BE CLOSED!

Thanks for bringing this up, I totally forgot this case. We started more discussion at cityofaustin/joplin#27

src/pages/Service.js
benweatherman: Totes! I just wasn't sure because we were assuming single location/contact in the prev code. Filed cityofaustin/janis#51 to make sure it gets fixed.

PR #44, opened 2017-12-18
Add live preview

benweatherman:
This gets compiled out in production (since @ifsimicoded rightly pointed out this could be an easy target for attackers), but will allow joplin to pass over URL params to set the correct data.

I've been testing this locally and I don't seem to run into any length limitations. If that happens we'll need to compress the data, but I don't want to do that before it's necessary.

### DIFF
--- a/src/pages/Service.js
+++ b/src/pages/Service.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import { get } from 'lodash';
 import axios from 'axios';
+import { parse } from 'query-string';
 
 import ContentItems from 'components/ContentItems';
 import RelatedLinks from 'components/RelatedLinks';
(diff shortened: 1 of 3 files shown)


src/pages/Service.js
```diff
@@ -20,6 +21,16 @@ class Service extends Component {
 
 
   componentDidMount() {
+    if (process.env.NODE_ENV !== 'production') {
+      // Allow querystrings to set data, which is used in joplin for livepreview
+      const query = parse(this.props.location.search);
```
ifsimicoded: Live preview currently seems very coupled to the service page. Any plans for how we might achieve preview for all pages?
Also, would postMessage work for passing content to the iframe in a cleaner/safer way?
https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

ifsimicoded:
giphy-downsized-large

src/pages/Service.js
benweatherman: It's definitely coupled. I'm honestly not sure the best way to get this working for all pages. Googling around shows some examples of listening to react router and then setting component props. I'm open to any suggestions on how this should fit in with the app.

I considered postmessage for this but I wasn't sure where a good place to put a listener would be. Since the child doesn't (currently) need to pass data to the parent, I decided to go the react router way. postmessage might make things easier regarding the "how does this work for all pages" question. I think either way is safe since it won't be active on a resident-facing site.

TL;DR I am terrible at javascript.

PR #45, opened 2017-12-19
Deploy to heroku

benweatherman:
Add a script for deploying to heroku

Some other tweaks
- Don't ignore any files for docker because explicitly allowing certain folders seems to confuse some tools
- Modify Dockerfile to accept `PORT` environment variable, which gets set by heroku
- Change instructions for adding packages so it's 1 command instead of several
- Use the longer form for flags in bash so it's easier to read

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,0 @@
-*
-!public/
-!src/
-!package.json
-!yarn.lock
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,17 @@
-from node
+FROM node
+
 RUN mkdir /app
 WORKDIR /app
-ENV PORT=80
+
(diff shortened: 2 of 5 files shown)


benweatherman:
Young Brent is astutely checking out the new deployment at https://janis-staging.herokuapp.com/services

ifsimicoded:
Can you explain more about the tools that confused when you ignore certain folders for my own edification?

benweatherman:
The `now.sh` commandline tool would ignore all files (I think because of the first line). I never followed up on filing an issue or investigating.

I also thought I was having trouble with the heroku CLI but now that I'm trying to reproduce it, it seems to work correctly.

Since it was just user error, I'll revert that change 😹

PR #47, opened 2017-12-21
Menu bar

### LINKED ISSUE #21: Header/Footer
Make header and footer components

- Header
  - [x] Language Bar
  - [x] Menu Bar
- [x] Footer

### LINKED ISSUE #32: Menu
Here's abbreviated mocks of the homepage with menu link and expanded menu
screen shot 2017-12-14 at 9 47 19 am
screen shot 2017-12-14 at 9 47 33 am

mateoclarke:
Closes #32 & Closes #21 

menu

I think this gets us most of the way there. We'll def have to refactor code around the navigation hierarchy. For example, what happens when you click a Service Type? We currently don't have an index of services by type. I think we should create an endpoint that gives us the data in a good format, but I made due with an existing endpoint. 

Other things we may to add would be an event lister on the esc key and on a click on the overlay area to trigger a state change and close the menu.

We may also need to check this for accessibility. Do we want tab indexing to skip this when its not visible?


src/App.js
```diff
@@ -22,11 +23,17 @@ class App extends Component {
     return (
       <Router>
         <div>
-          <Banner />
-          <I18nBanner />
+          <Route path="/" render={props => (
```
mateoclarke: in order to give the Header component `props.location`, I had to wrap it in a Route component. We'll see how this may need to change in an SSR world.

src/components/Navmenu.js
```diff
@@ -0,0 +1,122 @@
...
+
+  focusOnClose = () => {
+    this.refs.closeTrigger.focus()
+  }
+
+  componentDidMount () {
+    const endpoint = [
+      `${process.env.REACT_APP_CMS_ENDPOINT}/pages/`,
+      `?format=json&type=base.ServicePage&fields=topic(text)`,
+    ].join('')
```
mateoclarke: do y'all know if there a better syntax for a mutli-line string?

src/components/Navmenu.js
```diff
@@ -0,0 +1,122 @@
...
+            return data.topic.text === title
+          })
+        }
+
+        const nav = parentNavItems.map((title) => {
+        	return {
+        		parentTitle: title,
+        		children: getChildrenNavItemsByParent(title),
+        	}
+        })
```
mateoclarke: All this nonsense (lines 33-46) could be removed if we had a better endpoint.

src/components/Navmenu.js
```diff
@@ -0,0 +1,122 @@
...
+                <li>
+                  <a className={this.getParentMenuItemClassName(parent)} href="/services">
+                    {parent.parentTitle}
+                  </a>
+                  <ul className="usa-sidenav-sub_list">
+                  { parent.children.map((child) => {
+                    return (
+                      <li>
+                        <a className={this.getMenuItemClassName(`/service/${child.id}`)}
+                          href={`/service/${child.id}`}
```
mateoclarke: this will need to change if when we switch to slugs instead of ids in the routing

src/components/Navmenu.js
```diff
@@ -0,0 +1,122 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      nav: []
+    };
+  }
+
+  menuClassName = () => {
+    const baseClassName = `coa-Navmenu`;
+    return this.props.isOpen ? `${baseClassName}--open` : baseClassName;
```
ifsimicoded: For css convention — I was going with BEM class stacking (http://getbem.com/introduction/). So className=“base_class base_class—modifier”.  NOT just className=“base_class—modifier” 
Honestly, I haven’t spent a lot of time thinking/reading about what’s better here, just leaning on the folks who came up with BEM stuff to have done that heavy lifting.

src/components/Navmenu.js
```diff
@@ -0,0 +1,122 @@
...
+                      </li>
+                    )
+                  })}
+                  </ul>
+                </li>
+              )
+            })}
+          </ul>
+        </nav>
+        <div className={`coa-Navmenu__overlay${this.props.isOpen ? '' : '--hide'}`}></div>
```
ifsimicoded: I used the outline: 9999px solid rgba(0,0,0,0.5); trick on my modal overlay so I wouldn’t have to have multiple elements to create an overlay. Maybe this could work? Haven’t browser tested beyond chrome & safari desktop tested tho. Either way, maybe we can make generic overlay classes. Like coa-overlay, coa-overlay--center coa-overlay--left coa-overlay--right that can be used all over.

src/components/Header.js
```diff
@@ -1,77 +1,39 @@
...
-                <a href="/" title="Home" aria-label="Home">
-                  Welcome to Janis <span role="img" aria-label="peace sign">☮️</span>
-                </a>
-              </em>
+      <header className="coa-Header" role="banner">
+        <div className="wrapper">
+          <div className="row">
+            <div className="col-xs-6 coa-Header__menu">
+              <span onClick={this.toggleMenu} tabIndex="0">MENU</span>
+              <span className="coa-Header__text-spacer">|</span>
```
ifsimicoded: Can the separator ‘|’ be a part of a :after content property so it’s not read as content? Or do we want it to be read as content?
Can we also make this more generic -- coa-text_spacer_vertical -- so we can use these all over? I have a feeling we'll have links listed like this in other places.

src/components/Navmenu.js
```diff
@@ -0,0 +1,122 @@
...
+              return (
+                <li>
+                  <a className={this.getParentMenuItemClassName(parent)} href="/services">
+                    {parent.parentTitle}
+                  </a>
+                  <ul className="usa-sidenav-sub_list">
+                  { parent.children.map((child) => {
+                    return (
+                      <li>
+                        <a className={this.getMenuItemClassName(`/service/${child.id}`)}
```
ifsimicoded: Cursor pointer on the menu links plZZZZZ

src/App.js
```diff
@@ -22,11 +23,17 @@ class App extends Component {
...
+            <section>
+              <Banner />
+              <I18nBanner />
```
ifsimicoded: Looks like the translation bar should always be fixed over the menu. This way the menu slide is also consistently positioned over where a user initially clicked to reveal the menu. Can we throw some CSS at this?

src/components/Navmenu.js
```diff
@@ -0,0 +1,122 @@
...
+    const baseClassName = `coa-Navmenu`;
+    return this.props.isOpen ? `${baseClassName}--open` : baseClassName;
+  }
+
+  focusOnClose = () => {
+    this.refs.closeTrigger.focus()
+  }
+
+  componentDidMount () {
+    const endpoint = [
```
ifsimicoded: thinking we should start defining our endpoint urls in one place, since these will likely be changing and we’ll likely only want to update these urls in one place.

src/App.js
mateoclarke: > Looks like the translation bar should always be fixed over the menu. 
Are you referring to the mockups? Because I saw inconsistencies in how sidebars get displayed in them. I don't see a good reason functionally (or visually) to always show the language bar. Ok if we ask @cthibaultatx for clarification on this one?

src/components/Header.js
mateoclarke: sounds good, putting this in `helpers.scss`. Is that a good place for this type of thing?

src/components/Navmenu.js
mateoclarke: yeah, I don't really have a strong opinion here. I updated for this case, but unless there is a clear benefit, i'd leave it to chef's choice.

src/components/Navmenu.js
mateoclarke: cool, i'll create a file at `src/constants/endpoints` and import the string from there.

src/components/Navmenu.js
mateoclarke: i'm pretty sure we have that. it was missing on the "Menu" text so I added it there. Maybe that's what you meant.

src/components/Navmenu.js
mateoclarke: yeah, i just tried mimic what I saw the USWDS `usa-overlay` does. I was assuming it had good accessibility and browser coverage.

I'm open to refactoring this to for more holistic use.

mateoclarke:
@ifsimicoded, thanks for the feedback!

At first I was just going to leave the suboptimal jarring redirects until we land on what routing library we use with SSR. But when I looked into it, it wasn't too much to add. You can see what I did in the last commit. 
https://github.com/cityofaustin/janis/pull/47/commits/fa27bca92aa1ab75885f1d2d7450a553d231342f

I wasn't sure what you meant by...
> Can we create links via names, so that if routes change, we only have to make updates to links in one place? I have a feeling we'll be changing these.

ifsimicoded:
Hey when I said named routes, I meant like this: https://www.npmjs.com/package/react-router-named-routes
This way if we ever change the route for say the services page from /service/:id ... we could do it in one place, instead of updating links everywhere. Just something to think about it.

mateoclarke:
ah, gotcha. yeah, I think in an ideal world, we get back slugs/paths instead of (or in addition to) ids to use for these links.

I'll think we'll have to refactor this nav code regardless once the information architecture gets more solidified. This came up yesterday briefly in a chat with @benweatherman  & @cthibaultatx around page hierarchy. sounds like it'll be 3-4 levels of nesting.

Maybe we should anticipate having a Nav/menu content model in the CMS. What does Wagtail already give us there @benweatherman?

benweatherman:
> we get back slugs/paths instead of (or in addition to) ids to use for these links.

I think this is implementing cityofaustin/joplin#26

You can do this now via graphql (though it might change slightly)

```
{
  pages(slug:"dropoff") {
    edges {
      node {
        id
        content
      }
    }
  }
}
```

> Maybe we should anticipate having a Nav/menu content model in the CMS. What does Wagtail already give us there @benweatherman?

We'd just need to develop a content model to allow authors to create the hierarchy. If we're not ready for authors to do that yet, we could generate the data via the API like the client is currently doing without allowing authors to change it until we get around to developing the model. Filed cityofaustin/joplin#49 as a placeholder. It's not currently set for MVP, but let me know if that's not right.

PR #48, opened 2017-12-23
Sd cssbreakdown

### LINKED ISSUE #12: Component and CSS file/directory structure
Before we go too far, let's get all of our components organized in a good directory structure.

Other SASS things to think about:
- react component dir structure should probably be coherent with our sass dir structure
- package.json references flexboxgrid-sass git repo. is this risky? Not sure if pulling a package this way considers version changes
- Should we update all units to rems?
- update uswds variables with only variables we override

ifsimicoded:
Component and Sass Directory structure.
Basing this off a modified 7-1 Pattern --ish. http://vanseodesign.com/css/sass-directory-structures/
There's a lot of file location changes... let me know if this is a problem and I can work that out in a separate branch.
This will issue closes #12 .


src/components/modules/WYSIWYG.js
```diff
@@ -0,0 +1,15 @@
+import React, { Component } from 'react';
+
+class WYSIWYG extends Component {
```
mateoclarke: This component name was confusing for me. I was asking myself, why are we rendering an WYSIWYG editor on the public page? Until I realized the purpose of this is to inject HTML from that API that originally comes from the WYSIWYG editor in Joplin. Maybe we call it `HtmlFromAPI` or `HtmlFromAdmin`? idk. just thoughts.

PR #49, opened 2017-12-27
Service Page updates

### LINKED ISSUE #41: Update 'related services' section
We observed a handful of residents using the new digital services. One person referred to 'related services' and the 'feedback request form' as the _footer_. Another person using a tablet did not scroll down to the 'related services' section and so was never able to use it to discover and access similar services. Problem is that users are not recognizing 'related services' as part of the service page experience–it feels separate, like a footer. And on our end–'related services' is the primary next step we want users to take.

I imagined (A) moving the 'related service' content into the right column, beneath 'contact', but it felt like too much information. Of course, we can't be sure until residents test, but instead–let's (B) change the attributes of the 'related services' section so that when a user gets to the end of a service- they continue to scroll down and read over 'related services', believing that it is an inherent part of that service.

updates in mocks:
- change 'related services' section background to match service content background
- change service tile color to #0A71AE $color-primary
- change service tile type and chevrons to #ffffff white
- change 'related services' section title to: "Check out related city services"
**there's a change in column layout included in the mocks but I'll detail that in a separate issue**

mvp_l_1 1_service_15dec2017
mvp_m_1 1_service_15dec2017
mvp_s_1 1_service_15dec2017

ifsimicoded:
update 311 links with outbound urls
update contact section to handle multiple contacts (rely on stub data as Joplin model isn't updated)
update related service links based on Celine's mocks closes #41


src/__tmpdata/services.json
```diff
@@ -8,27 +8,107 @@
...
+      "phone":"512-367-8695",
+      "email":"resourcerecovery@austintexas.gov",
+      "address": {
+        "street": "2514 Business Center Drive",
+        "city": "Austin",
+        "state": "Texas",
+        "zip": "78744",
+        "country": "United States"
+      },
+      "hours": [
```
benweatherman: There's a slightly different structure for contacts & locations currently https://github.com/cityofaustin/joplin/issues/39#issuecomment-354149336. Some goofiness is there because of the default Django rest framework structure, but I separated contact & location. Not sure if that's the right way to model it or not.

ifsimicoded: made some comments here -- https://github.com/cityofaustin/joplin/issues/39
but basically, since there will be multiple locations and points of contact, I think there will be cases where these two should be associated when shown to residents. For ex, contact info for Austin Resource Recovery main drop off location vs Treehouse where classes are hosted.

src/components/layout/Service311.js
```diff
@@ -18,19 +18,17 @@ class Service311 extends Component {
...
-                  url="#"
+                  url={service.url}
                   text={service.title}
                 />
               </div>
             )
           }
           </div>
-          <a className="coa-section__link" href="#">See a Full List of 311 Services</a>
+          <a className="coa-section__link" href="http://311.austintexas.gov/reports/list_services">See a Full List of 311 Services</a>
```
mateoclarke: Should we do `target="_blank"` on these since we link them to a separate site? Or since it is the same domain, maybe note?

src/components/modules/Hours.js
```diff
@@ -22,7 +23,7 @@ class Hours extends Component {
             hours.map((hour, index) =>
               <tr key={index}>
                 <th scope="row">{hour.day_of_week}</th>
-                <td>{hour.start_time} - {hour.end_time}</td>
+                <td>{moment(hour.start_time, "HH:mm:ss").format('h:mm A')} - {moment(hour.end_time, "HH:mm:ss").format('h:mm A')}</td>
```
mateoclarke: 👍

src/components/layout/Service311.js
ifsimicoded: Not 100% sure, I can ask Celine what her thoughts are

PR #52, opened 2017-12-28
Match API changes for contact

benweatherman:
Updated URL and data to accomodate the structure returned by the API


benweatherman:
🚨This should probably just be closed since #55 is probably the way we'll go forward 🚨

src/pages/Service.js
```diff
@@ -32,13 +32,23 @@ class Service extends Component {
...
     axios
-      .get(`${process.env.REACT_APP_CMS_ENDPOINT}/pages/${this.props.match.params.id}?fields=content,extra_content,topic(text),locations(location(name,street,city,state,zip,country,hours)),contacts(contact(name,email,phone))`)
+      .get(`${process.env.REACT_APP_CMS_ENDPOINT}/pages/${this.props.match.params.id}?fields=content,extra_content,topic(text),contacts(contact(name,email,phone,hours,location(name,street,city,state,zip,country)))`)
       .then(res => {
         this.setState({ data: res.data })
       })
       .catch(err => console.log(err))
   }
 
+  cleanContacts(contacts) {
```
ifsimicoded: no need to add an additional method. We can just update address to location in the JSX and update Contact.js to reference contact.contact. I can make this update in another branch. Or also graphql!

benweatherman: I was trying to keep the sub-components from changing, but up to y'all how to structure stuff. Do you want me to make the change in the JSX for the graphql?

benweatherman:
I'm gonna close in lieu of the graphql version

PR #55, opened 2018-01-02
Fetch data from graphql endpoint

benweatherman:
live-preview

### DIFF
--- a/src/components/modules/Hours.js
+++ b/src/components/modules/Hours.js
@@ -22,8 +22,8 @@ class Hours extends Component {
           {
             hours.map((hour, index) =>
               <tr key={index}>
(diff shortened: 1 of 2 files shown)


benweatherman:
I'll go ahead and merge this just so we can get the graphql train rolling.

Did you have thoughts on where the query files might live? I've seen lots of variation. I kinda like having a `queries` directory at the same level as `components` or `pages` but might also be good to keep it closer to the thing that uses the query until we get more uses.

mateoclarke:
no strong opinions. same level as components or pages works for me

benweatherman:
Updated in cityofaustin/janis#72

PR #72, opened 2018-01-05
Update graphql usage

benweatherman:
- switch to navigating via slugs instead of IDs
- change `/services` to use graphql
- moved queries into own files to make the JS easier to read
- updated single page query to pass in variables
- API naming is changing slightly to make it more like other graphql APIs

Look! It works!
wonks

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -34,7 +34,7 @@ class App extends Component {
             <Route exact path="/" component={HomePage} />
             <Route exact path="/services" component={ServicesIndex} />
             <Route exact path="/search" component={SearchPage} />
-            <Route path="/service/:id" component={Service} />
+            <Route path="/service/:slug" component={Service} />
           </section>
           <Footer />
         </div>
--- a/src/pages/Service.js
+++ b/src/pages/Service.js
@@ -74,7 +74,7 @@ class Service extends Component {
     const steps = get(data, "content", null);
     const contentItems = get(data, "extraContent", null);
     const contacts = get(data, "contacts.edges", []).map((n) => this.cleanContact(n.node.contact));
-    const relatedlinks = get(jsonFileData, "servicesRelated", null);
+    const relatedlinks = get(data, "related", null);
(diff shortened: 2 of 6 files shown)


src/App.js
```diff
@@ -34,7 +34,7 @@ class App extends Component {
             <Route exact path="/" component={HomePage} />
             <Route exact path="/services" component={ServicesIndex} />
             <Route exact path="/search" component={SearchPage} />
-            <Route path="/service/:id" component={Service} />
+            <Route path="/service/:slug" component={Service} />
```
mateoclarke: 🐌

src/pages/Service.js
```diff
@@ -9,6 +9,7 @@ import RelatedLinks from 'components/layout/RelatedLinks';
 import FormFeedback from 'components/layout/FormFeedback';
 import Service311 from 'components/layout/Service311';
 import WYSIWYG from 'components/modules/WYSIWYG';
+import servicePageQuery from 'queries/servicePageQuery';
```
mateoclarke: thanks for moving this into a queries folder. very tidy 💅

src/pages/Services.js
```diff
@@ -19,9 +20,11 @@ class ServicesIndex extends Component {
 
   componentDidMount() {
     axios
-      .get(`${process.env.REACT_APP_CMS_ENDPOINT}/pages/?type=base.ServicePage`)
+      .post(`${process.env.REACT_APP_CMS_ENDPOINT}/graphql/`, {
+        query: allServicePagesQuery,
+      })
       .then(res => {
-        this.setState({ data: res.data })
+        this.setState({ data: res.data.data.allServicePages })
```
mateoclarke: is `data.data` code smell? 🧀

benweatherman: ugh, it's terrible. `axios` uses `.data` as "the data that's returned in the response" and graphql's default thing is to return a `data` key. I could do more destructuring somewhere else but we're kinda stuck with `.data.data` somehow.

PR #74, opened 2018-01-05
React Intl

mateoclarke:
Adds integration of React-Intl with React-Router v4. `react-intl` acts as a store holding onto language state. We could use other conveniences it provides around referencing language libraries (static files with translations), formatting for numbers/currency, and pluralization. 

Right now, when the user clicks on the Language select, we reroute their url path, fetch new data nd update components.

jan-09-2018 01-54-50


src/js/components/LanguageWrapper.js
```diff
@@ -0,0 +1,82 @@
...
+
+import SUPPORTED_LANGUAGES from 'js/constants/languages'
+
+class LanguageWrapper extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      lang: this.setLanguage()
+    }
+    this.daysUntilCookieExpires = 7
```
benweatherman: It seems like we'd want to keep it around 4eva

src/js/pages/Service.js
```diff
@@ -58,6 +52,9 @@ class Service extends Component {
     }
 
     axios
+      .create({
+        headers: {'Accept-Language': lang }
```
benweatherman: Your end `}` has a space before it but not after the leading `{`

I'm sorry to even point this out, but I saw it so...

src/js/pages/Services.js
```diff
@@ -20,6 +20,9 @@ class Services extends Component {
 
   componentDidMount() {
     axios
+      .create({
+        headers: {'Accept-Language': this.props.lang }
```
benweatherman: This has that unbalanced close `}` thingy

src/js/page_sections/I18nBanner.js
```diff
@@ -1,50 +1,34 @@
...
-    this.visibleLanguageOptions = languages.slice(0, this.displayedLanguageAmount);
-    this.secondaryLanguageOptions = languages.slice(this.displayedLanguageAmount, languages.length);
+    this.visibleLanguageOptions = SUPPORTED_LANGUAGES.slice(0, this.displayedLanguageAmount);
+    this.secondaryLanguageOptions = SUPPORTED_LANGUAGES.slice(this.displayedLanguageAmount, SUPPORTED_LANGUAGES.length);
+    console.log(this.getPathname())
+  }
+
+  getPathname = () => {
+    let pathArray = this.props.location.pathname.split('/').filter(n => n)
+    if (SUPPORTED_LANGUAGES.map(lang => lang.code).includes(pathArray[0])){
```
benweatherman: How would you feel about pre-calculating `SUPPORTED_LANG_CODES` inside of `constants.js`. It would save the recalculation in other places, like https://github.com/cityofaustin/janis/pull/74/files#diff-05c601d58da6bea7d129c7ef04b9e2feR33

src/js/pages/Service.js
mateoclarke: stop trying to crush my individuality

src/js/page_sections/I18nBanner.js
mateoclarke: i like it!

src/js/pages/Service.js
benweatherman: Enjoy yourself spidey!

image

PR #75, opened 2018-01-05
Sd issue12issue29 dirstructure

ifsimicoded:
Celine's handling the topic page in a different branch!!!!

This branch closes out issue #12 and feedback from the last dir structure PR, though this will probably be an ongoing work. 
It's a lot of file changes, but again, it's mostly renames.


README.md
```diff
@@ -25,13 +25,33 @@ docker exec --interactive --tty janis yarn add <package name>
 
 ---
 
+## SASS Regime Guidelines (WIP)
```
mateoclarke: 🔨 👊

README.md
```diff
@@ -25,13 +25,33 @@ docker exec --interactive --tty janis yarn add <package name>
...
+- our css is namespaced with coa-
+- our css classes should stack. EX. a blue button would look like `<button class="coa-button coa-button--blue">Button</button>`.
+- class names for js components should correspond with the js component name(capitalized and camelCased).
+  EX. LinkList.js markup has styles applied via the class names `coa-LinkList coa-LinkList--boxprimary`.
+- class names which are not js components but have multiple words should be separated by a -
+  EX. `coa-Footer__body-text`
+
+Resources:
+- http://vanseodesign.com/css/sass-directory-structures/
+- http://getbem.com/introduction/
```
mateoclarke: 👏

README.md
```diff
@@ -25,13 +25,33 @@ docker exec --interactive --tty janis yarn add <package name>
...
+
 ## Accessibility Guidelines (WIP)
 
-- When we use `target="_blank"` on an anchor tag, we should add `aria-label="Opens in new window"`
+- When we use `target="_blank"` on an anchor tag, we should add `aria-label="Opens in new window"` and `rel="noopener noreferrer"` for security
 
 a11y Resources:
 - https://accessibility.18f.gov/
 - https://cfpb.github.io/design-manual/best-practices/accessibility-best-practices.html
+- https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/
```
mateoclarke: This is nit picky, but this isn't really an a11y resource. Maybe we just use markdown to link `[link](url)` this article above inline with where we talk it. Also, since this vulnerability gets linted by React, not sure how important it is to call out here, though not against it in principle.

README.md
```diff
@@ -25,13 +25,33 @@ docker exec --interactive --tty janis yarn add <package name>
 
 ---
 
+## SASS Regime Guidelines (WIP)
+
+We're using a modified form of the 7-1 pattern, and BEM for CSS naming and organization
+- our css file structure and js file structure should be congruent.
+  EX. You can find css for js modules in the css/modules folder, you can find css for js page_sections in the css/page_sections folder and so on.
+- js/modules our are smallest unit and should not have any additional custom react component imports
```
mateoclarke: I'm not opposed to this strict definition of components in js/modules as the smallest atomic unit with no child components. But where do we put components that have children but aren't page sections.

ifsimicoded: you know, I haven't thought this through that far, but definitely open to adding another folder called components when we need it?

README.md
ifsimicoded: *nod*

README.md
mateoclarke: cool, i think its coming in my next PR and I'll take that suggestion as your blessing. If we wanted to align with Reactisms, we might consider the delineation not only if the component renders child components but also if the component has state or not. `js/modules` would be functional stateless components and `js/components` would be stateful components or components that render other components.

https://medium.com/@juanguardado/stateful-vs-stateless-components-444b5aa21865

PR #77, opened 2018-01-08
Sd navmenu usegraphql

ifsimicoded:
- Updated Nav Menu to use graphql endpoint to fetch service links by topic for all topics.
- Fixed nav menu/service page bug which rendered service page before new data was fetched.

Branch off of #75


src/js/page_sections/Navmenu.js
```diff
@@ -19,104 +20,109 @@ class Navmenu extends Component {
...
+
+  getMenuItemClassName = (path) => {
+    const currentPath = this.getCurrentPath();
+    return currentPath === path ? 'usa-current' : '';
   }
 
   render() {
-    return (
+
+    if (!this.state.isLoaded) return 'loading...';
```
mateoclarke: We might want to ask Celine is they have a preferred loading state. But then again, this won't matter after we go SSR or SSG so this implementation is 🆗

src/js/page_sections/Navmenu.js
```diff
@@ -19,104 +20,109 @@ class Navmenu extends Component {
...
+
+    const { edges: parentLinks = [] } = this.state.data;
+
+    return parentLinks.length && (
       <div className="usa-grid-full">
-        <nav role="navigation" className={this.menuClassName()}>
+        <nav className={this.menuClassName()}>
```
mateoclarke: Not questioning that is the right thing. But please educate me on why we should remove the `role="navigation"` here.

src/js/page_sections/Navmenu.js
```diff
@@ -19,104 +20,109 @@ class Navmenu extends Component {
...
+
+            let { edges: serviceLinks = [] } = parentLink.services;
+
+            return (
+              <li key={parentLink.id}>
+                <Link to="/services" className={this.getParentMenuItemClassName(parentLink)}>
+                  { parentLink.text }
+                </Link>
+
+                { serviceLinks.length && (
```
mateoclarke: Should we stay consistent with naming and call this `childLinks` for `parentLinks`?

mateoclarke: instead of `serviceLinks`?

src/js/page_sections/Navmenu.js
ifsimicoded: This was just a linter fix. I think nav elements has an implicit navigation role.
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-redundant-roles.md

src/js/page_sections/Navmenu.js
ifsimicoded: I kept this specific, since the API returns specifically service links and in the future might return dept links and/or event links etc. We're also currently being specific on how we build paths (/service/slug/) -- so I figured specifying would be best since event and dept links would be built differently.

Issue #33, opened 2017-12-14
Implement homepage design


cthibaultatx:
- [x] update tile/service and module/services in sm, med, and lge

- [x] add 'work in progress' message in sm, med, and lge

- [ ] ~add theme to main navigation plus...~

- [ ] ~...refine mobile and tablet menu and develop new large/desktop menu~

- [ ] ~Themes and topics are listed here: [Future Navigation airtable](https://airtable.com/shr3RatfCvSZmJhqw) Only build in topics we already have services/content for ('Recycling, Trash, and Compost' under 'Housing & Utilities'). Leave off other topics and instead, when a user hovers over a theme with no topics- they should read: (waiting on copy from content team @courtneyjacinic )~

- [ ] ~add airport icon to tablet and desktop header, and in mobile menu- links to http://austintexas.gov/department/airport~

- [ ] ~add 311 link to header to tablet and desktop, and in mobile menu- links to http://311.austintexas.gov/reports/list_services~

- [x] ~add 'alpha' to 'austin.gov' = 'alpha.austin.gov'~

## Do not change these things:
- language header
- header/footer colors
- 311 and footer content - will be separate issue
- typestyles

![alpha_l_0 0_home_menua](https://user-images.githubusercontent.com/20911665/36178991-5141132a-10e0-11e8-9f7b-ac40a9b02b17.png)

![alpha_l_0 0_home_menub](https://user-images.githubusercontent.com/20911665/36178992-5154bc7c-10e0-11e8-9a3a-76c74105410b.png)

![alpha_l_0 0_home](https://user-images.githubusercontent.com/20911665/36178993-516a44a2-10e0-11e8-9815-1b68c8160c32.png)

![alpha_m_0 0_home_menu](https://user-images.githubusercontent.com/20911665/36178994-517d4d22-10e0-11e8-8d3a-97254cbc42a5.png)

![alpha_m_0 0_home](https://user-images.githubusercontent.com/20911665/36178995-5190cf50-10e0-11e8-8e63-cec6f5f84626.png)

![alpha_s_0 0_home_menu](https://user-images.githubusercontent.com/20911665/36178996-51a5d832-10e0-11e8-9fae-36cb09da41eb.png)

![alpha_s_0 0_home](https://user-images.githubusercontent.com/20911665/36178997-51bc0166-10e0-11e8-901e-ed456f060121.png)

[...]

PR #80, opened 2018-01-09
Updated home, added USWDS Search module

cthibaultatx:
This is the start of editing the homepage #33

### DIFF
--- a/src/js/modules/GlobalSearch.js
+++ b/src/js/modules/GlobalSearch.js
@@ -1,0 +1,23 @@
+import React, {Component} from 'react';
+
+class GlobalSearch extends Component {
+
+  render() {
+
+    return  (
+
+      <div>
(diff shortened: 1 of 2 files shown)


src/js/modules/GlobalSearch.js
```diff
@@ -0,0 +1,23 @@
+import React, {Component} from 'react';
+
+class GlobalSearch extends Component {
+
+  render() {
+
+    return  (
+
+      <div>
```
ifsimicoded: super minor for accessibility, but USWDS has a role on the wrapper div.
`<div role="search"><form class="usa-search usa-search-big">...</form></div>`

mateoclarke: Let me know if you need any help with this @cthibaultatx

PR #81, opened 2018-01-10
add topic page, query and update topic page links

ifsimicoded:
add topic page, query and update topic page links
We'll probably need to update these pages to have slugs and load from slugs

### DIFF
--- a/src/js/pages/Topic.js
+++ b/src/js/pages/Topic.js
@@ -35,14 +35,12 @@ class Topic extends Component {
   shouldComponentUpdate(nextProps, nextState) {
     // only rerender component when state has changed
     // this happens only when data is refetched
-console.log(this.state.data.id, nextState.data.id);
     if (nextState.data.id !== this.state.data.id) return true;
 
     return false;
   }
 
   fetchData(id) {
-
     axios
       .post(`${process.env.REACT_APP_CMS_ENDPOINT}/graphql/`, {
         query: topicPageQuery,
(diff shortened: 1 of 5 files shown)


src/js/queries/topicPageQuery.js
```diff
@@ -0,0 +1,23 @@
+const topicPageQuery = `
+  query topicPageQuery($pk:ID) {
```
benweatherman: Small nit: we should call it `$id` instead of `$pk`. The latter is used in `servicePageQuery` to aid the transition from database ID to slug (and should probably just be removed now). In this case we're using graphql IDs, so I think it'll be less confusing.

src/js/pages/Topic.js
```diff
@@ -0,0 +1,111 @@
...
+
+  shouldComponentUpdate(nextProps, nextState) {
+    // only rerender component when state has changed
+    // this happens only when data is refetched
+    if (nextState.data.id !== this.state.data.id) return true;
+
+    return false;
+  }
+
+  fetchData(id) {
```
mateoclarke: We're using this method in a couple different Page Components. And it seems like the `componentDidMount` and `shouldComponentUpdate` patterns are pretty consistent too. Not saying now is the time, but should we maybe think about a uniform abstraction for fetching data from the GraphQL server?

mateoclarke: My first instinct was a utility function, but i'd be open to seeing how other React-y patterns my be useful here.

Issue #34, opened 2017-12-14
Department page template


cthibaultatx:
- [ ] Leave out project tracker section

![mvp_s_2 0_dept](https://user-images.githubusercontent.com/20911665/34007369-26e7d0ee-e0c7-11e7-8b3a-7c369c6ba20c.png)

ifsimicoded:
Left in the project tracker section, but it can easily be commented out/removed. 
Closed by #83

ifsimicoded:
Left in the project tracker section, but it can easily be commented out/removed.
Closed by #83

PR #83, opened 2018-01-11
Department Page + stuff

ifsimicoded:
- Department page
- hide sections we don't have data for
- contact module update so cleaning happens once (thinkin out contacts data will be formatted the same regardless of what page we get it on)
- listlink module update so we can reuse


src/js/pages/Department.js
```diff
@@ -0,0 +1,119 @@
...
+            </div>
+          </div>
+        </div>
+
+        <RelatedLinks
+          relatedlinks={relatedlinks}
+          sectionType="secondary"
+          sectionLink={{url: "#", text: "Track all Resource Recovery projects"}}
+          sectionTitle="Track Resource Recovery Projects"
+          sectionText="Projects are short term, with a set budget, and defined goals. Projects can be specific to one department or a collaboration across multiple departments."
```
benweatherman: 🤓

src/js/page_sections/Contact.js
```diff
@@ -6,61 +6,76 @@ import Hours from 'js/modules/Hours';
...
-          })
-        }
-      </div>
-    );
+    let JSX;
+
+    if (!contacts || !contacts.length) {
+      JSX = null;
+    } else {
+      JSX = (
```
mateoclarke: this is a nifty pattern I hadn't really seen before

src/js/modules/ListLink.js
```diff
@@ -6,12 +6,11 @@ class ListLink extends Component {
...
   render() {
 
-    const { id, url, text, isBoxType } = this.props;
+    const { url, text, linkStyle } = this.props;
 
     return (
       <Link
-        className={ `coa-ListLink ${(isBoxType && "coa-ListLink--boxprimary" )}` }
-        key={id}
+        className={ `coa-ListLink coa-ListLink--${(linkStyle)}` }
```
mateoclarke: I like how you pass the modifier into classname right from the props. This makes me think that we should get back to adding PropTypes in our components. It will help us avoid a situation where this modifier gets an unexpected value. Should I add that to the backlog?

ifsimicoded: yes plz.

src/js/page_sections/Contact.js
ifsimicoded: it's my way of getting around linter requirements that always expect return statements.

src/js/helpers/cleanData.js
```diff
@@ -0,0 +1,38 @@
...
+  });
+};
+
+export const cleanRelatedServiceLinks = (links) => {
+
+  if (!links) return null;
+
+  return links.map((link) => {
+    return {
+      url: getPathWithLangCode(`/service/${link.slug}`),
```
ifsimicoded: removed the getPathWithLangCode from some react modules since not all links are inbound links. It was goofy with 311 outbound links (see screenshot)
screen shot 2018-01-12 at 4 46 06 pm

This will modify our url data during a cleanData step after data is fetched.

src/js/modules/ListLink.js
```diff
@@ -6,13 +6,12 @@ class ListLink extends Component {
...
   render() {
 
-    const { id, url, text, isBoxType } = this.props;
+    const { url, text, linkStyle } = this.props;
 
     return (
       <Link
-        className={ `coa-ListLink ${(isBoxType && "coa-ListLink--boxprimary" )}` }
-        key={id}
-        to={getPathWithLangCode(url)}
```
ifsimicoded: Removed getPathWithLang because the url data has already been parsed when data is fetched.

src/js/page_sections/RelatedLinks.js
```diff
@@ -8,37 +8,55 @@ class RelatedLinks extends Component {
...
-    const { relatedlinks, topicId, topicName } = this.props;
-
-    return (relatedlinks) && (
-
-      <div className="coa-section">
-        <div className="wrapper">
-          <SectionTitle title="Check out related city services" noBorder={true} />
-
-          <div className="row">
+    const { relatedlinks, sectionStyle, sectionTitle, sectionText, sectionLink } = this.props;
```
ifsimicoded: made this page_section more generic since we have a lot of places now rendering lists of box links.

src/js/pages/Service.js
```diff
@@ -64,19 +65,17 @@ class Service extends Component {
         }
       })
       .then(res => {
-        this.setState({ data: res.data.data.servicePage });
+        const data = this.cleanData(res);
```
ifsimicoded: if we cleanData (get it in the right format) before it's thrown in the state, this will prevent the issues we were seeing where the app would bomb trying to clean already cleaned data in the render method. Basically, this replaces the logic in 'shouldComponentUpdate'

ifsimicoded:
closes #34

PR #84, opened 2018-01-12
React intl

mateoclarke:
This got merged into the wrong branch originally. Retrying what should have been #74


src/js/pages/Service.js
```diff
@@ -27,25 +27,19 @@ class Service extends Component {
     this.fetchData(this.props.match.params.slug);
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentWillReceiveProps(nextProps, nextState) {
```
ifsimicoded: really feeling you on the more global solution for data fetching.

PR #85, opened 2018-01-12
Use react router Link component

mateoclarke:
This PR is doing basic stuff. Replacing were we used `<a>` tags with the `<Link>` component provided by react-router-dom so we have that quick and smooth single page app style page transition instead of a page reload. 

When we switch to something SSR, that framework will likely have their own Link component provided with the routing library so we'll need to update which library we import Link from along with any other refactors required.

I also wrapped our Page Component routes in a `<Switch>` since they should all be mutually exclusive. We never render the Home Page and a Service Page together.


mateoclarke:
yeah, we should add an issue all around refocusing to the top element of the main section when the page reroutes.

PR #97, opened 2018-01-16
Add issue template

benweatherman:
It was suggested that having a template will make it easier to create issues. Let's try that out!


benweatherman:
I made a few more iterations and am going to merge as good enough for now. Please let me know if there are other changes we should make!

PR #104, opened 2018-01-18
Recollect

ifsimicoded:
recollect app integration
minor fix for section spacing


src/css/vendor/_overwrites_flexboxgrid.scss
```diff
@@ -0,0 +1,9 @@
+.wrapper {
+  max-width: $coa-container-maxwidth;
+  padding-left: $coa-space-level5;
+  padding-right: $coa-space-level5;
+  @include media($coa-xlarge-screen) {
```
mateoclarke: Why do we drop padding for xlarge screens?

src/js/vendor/Recollect.js
```diff
@@ -0,0 +1,63 @@
...
+
+  componentWillMount() {
+    // TODO: temp fix as current implementation of recollect app is not SPA friendly
+    if(window.Recollect) {
+      window.location.reload();
+    }
+  }
+
+  shouldComponentUpdate() {
+    return false;
```
mateoclarke: Do we need this still?

src/js/vendor/Recollect.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  render() {
+    return (
+        <div id="rCw" ref={el => this.el = el}>
+          <img alt="Loading" src="https://recollect.a.ssl.fastly.net/0.11.1516038288/images/loading.gif"/>
```
mateoclarke: Should we just download this as a local asset?

src/js/vendor/Recollect.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    return name;
+  }
+
+  handleScriptLoad() {
+
+    const name = this.getRecollectWidgetName();
+
+    let script = document.createElement('script');
```
mateoclarke: Could you maybe add some comments here to describe what happens here and why it's kind janky?

benweatherman:
@ifsimicoded `id=7` now has only 1 recollect block and it's using the new block stuff instead of `application_block`

src/js/vendor/Recollect.js
ifsimicoded: I didn't cause I figured Celine might have a different idea for what a loader should look like.

src/js/vendor/Recollect.js
ifsimicoded: This was recommended in the implementation docs for third party scripts in react so any parent updates wouldn't result in downstream repaints, which is different than a remount.
Really, none of these stateless components should be components -- they should just be functions that spat out JSX -- and this one would have a check and reload before it spat out the JSX. But that's a bigger refactor.

src/js/vendor/Recollect.js
ifsimicoded: yep

PR #107, opened 2018-01-19
Static Maps

ifsimicoded:
This gets maps for application blocks and contact on to the page and builds off the recollect PR/branch. There will be additional changes to props as Joplin is updated.

I went with Google Maps Static Api since it auto-geocodes and doesn't require any additional js libraries.
Usage is limited to: 25,000 map loads per day, which we can be conscious of as post MVP or as our map needs become more robust.


mateoclarke:
Hey, small thing, but could you add screen grabs to PRs for things like this in the future?

src/js/modules/StaticMap.js
```diff
@@ -0,0 +1,48 @@
...
+
+    if (street) {
+      return encodeURIComponent(`${street} ${city || "Austin"}, ${ state || "Texas"} ${zip}`);
+    }
+
+    if (latitude && longitude) {
+      return encodeURIComponent(`${latitude}, ${longitude}`);
+    }
+
+    return encodeURIComponent("City Hall, Austin, Texas 78701");
```
mateoclarke: Could we maybe save this string as a const called something like `fallbackAddress`? Or just something to make it clear in for others?

src/js/modules/StaticMap.js
```diff
@@ -0,0 +1,48 @@
+import React, {Component} from 'react';
+
+class StaticMap extends Component {
+
+  getLocation() {
+
+    const { street, city, state, zip, latitude, longitude } = this.props.location;
```
mateoclarke: Are we going to be getting lat and longs from Joplin?

mateoclarke: nvm, see you passing in some hardcoded lat/lons for other maps.

src/js/modules/StaticMap.js
ifsimicoded: yep

src/js/modules/StaticMap.js
ifsimicoded: yea, not entirely sure what we'll be getting so this is just flexible.

ifsimicoded:
Can definitely log the Google quota issue, but we either never hit it at Sparefoot or they were ultra flexible with us and we had a nutty amount of traffic.

PR #102, opened 2018-01-17
Navigation cleanup

ifsimicoded:
Closes #101 
implemented NavLinks component to simplify active class logic.

PR #109, opened 2018-01-20
Fix active state in Navmenu

mateoclarke:
Active state styling wasn't being applied in the Navmenu because of language code prefixed in the paths. 

Doing this made me think that the `NavLink` component provided by react-router-dom is a better solution than this. ex: #102  🙃 

screen shot 2018-01-19 at 10 34 32 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "react": "^16.1.1",
     "react-dom": "^16.1.1",
     "react-intl": "^2.4.0",
+    "react-load-script": "^0.0.6",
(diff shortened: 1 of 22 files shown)


mateoclarke:
@ifsimicoded said she reviewed this and told me to merge. We agreed this is a short term fix and the existing router functionality (NavLink, basename, etc) should be used once we choose a router for server-side/static generation.

We do as Simi says. 🛠 ☠️ 🏴

PR #115, opened 2018-01-24
google analytics implementation

### LINKED ISSUE #30: Plug and configure analytics
- Define actionable analytics using user behavior learnings from research and austintexas.gov existing analytics
- Plan configuration with goals and data collection
- Implement

ifsimicoded:
I've left this in debug mode, so we can see page views logging in console. Not 100% sure how this will interact with SSR since I'm not 100% on how our routing works or what we'll land on for routing.
I'll also be following this issue thread https://github.com/react-ga/react-ga/issues/210

closes #30 -- until we have better definition around what our logging needs are.

### DIFF
--- a/src/js/helpers/GoogleAnalyticsPageView.js
+++ b/src/js/helpers/GoogleAnalyticsPageView.js
@@ -1,0 +1,13 @@
+import ReactGA from 'react-ga'
+
+const GoogleAnalyticsPageView = () => {
+  ReactGA.set({
+    page: window.location.pathname,
+    location: `${window.location.origin}${window.location.pathname}${window.location.search}`
+  });
+
+  ReactGA.pageview(window.location.pathname);
+  return null;
+};
+
+export default GoogleAnalyticsPageView;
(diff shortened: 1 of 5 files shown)


src/js/helpers/GoogleAnalyticsPageView.js
```diff
@@ -0,0 +1,13 @@
+import ReactGA from 'react-ga'
+
+const GoogleAnalyticsPageView = () => {
+  ReactGA.set({
+    page: window.location.pathname,
+    location: `${window.location.origin}${window.location.pathname}${window.location.search}`
```
mateoclarke: since I'm a GA dummy, can you explain why we set a page and location separately?

ifsimicoded: Not a dummy at all -- these two values populate the dl and dp variables that are sent back to google, and I noticed that dl was not updating with pushState changes -- only with full page loads. The issue I'm following in the react-ga repo, I mentioned above, talks about this some.
Here's some info [GA params](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters)

Issue #79, opened 2018-01-09
Collapse contact hours and make expandable


cthibaultatx:
![screen shot 2018-01-09 at 8 55 47 am](https://user-images.githubusercontent.com/20911665/34726930-0af59fd4-f51b-11e7-93f7-3949554a146e.png)

- [ ] defaults to current day of the week plus corresponding hours

- [x] when expanded, shows days following meaning if it's Wednesday, the expanded list starts with Wednesday, ends with Tuesday

- [ ] once expanded, current day of week and hours is bolded to Open Sans semi-bold, weight 600

Not sure if the day should also be formatted in p-link style (highlighted and underlined in blue). Referencing google- they only link the hours, not the day.

## Resident Value
doesn't bombard resident with unrelated information and gives them control to see more if desired

ifsimicoded:
@cthibaultatx We've been using Montserrat for our san font face based on the mocks. Do we need to also use Open Sans? For each new font we load, the user has to download that font resource so limiting the number of fonts we use is critical for performance/speed-experience.

PR #118, opened 2018-01-25
Sort and format hours/days beginning with current day.

jessereitz:
#79 sort and displays hours. Still needs to collapse hours. Assumes Joplin provides data for all 7 days of the week.

### DIFF
--- a/src/js/helpers/cleanData.js
+++ b/src/js/helpers/cleanData.js
@@ -1,4 +1,5 @@
 import { getPathWithLangCode } from 'js/helpers/language';
+import moment from 'moment';
 
 export const cleanContacts = (contacts) => {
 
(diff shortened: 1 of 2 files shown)


src/js/helpers/cleanData.js
```diff
@@ -7,7 +8,15 @@ export const cleanContacts = (contacts) => {
   return contacts.edges.map(({node: contact}) => {
     let cleaned = Object.assign(contact.contact);
     if(cleaned.hours && cleaned.hours.edges) {
-      cleaned.hours = cleaned.hours.edges.map(({ node: hours }) => hours);
+      cleaned.hours = cleaned.hours.edges.map(({ node: hours }) => {
+        hours.startTime = moment(hours.startTime, "HH:mm:ss").format('h:mm A');
+        hours.endTime = moment(hours.endTime, "HH:mm:ss").format('h:mm A');
+        let dayMoment = moment(hours.dayOfWeek, 'dddd');
+        hours.dayOfWeek = dayMoment.format('dddd');
+        hours.day = dayMoment.format('E');
```
mateoclarke: This is really small, but why not get rid of line 14 and just let line 16 be: 
`hours.day = moment(hours.dayOfWeek, 'dddd').format('E');`

That `let` definition in the middle of the rest kinda threw me off and you're already chaining the `.format()` function on all the other lines.

src/js/modules/Hours.js
```diff
@@ -1,11 +1,19 @@
 import React, {Component} from 'react';
 import moment from 'moment';
+import {sortBy, findIndex} from 'lodash';
 
 class Hours extends Component {
 
-  render() {
+  sort() {
+    // TODO: Joplin data MUST include data for all 7 days of week.
```
mateoclarke: I'm absolutely cool with leaving in TODO comments like this, but just want to make sure there is an Issue created in the Joplin repo for this.

mateoclarke:
@jessereitz, I removed "Fix" from your PR comment just so that it doesn't automatically close the issue before we get to the collapsing part. Thanks for tagging this PR back to that issue though!

src/js/modules/Hours.js
ifsimicoded: Yep! https://github.com/cityofaustin/joplin/issues/27

jessereitz:
@mateoclarke, sorry it's taken so long! Removing "fix" is absolutely, 100% okay with me! I'll take a look into actually getting that fix sorted

PR #119, opened 2018-01-26
React Static

mateoclarke:
_Completes #5_

With this PR, we can now export our react app as a static site. We do this using [React Static](https://github.com/nozzle/react-static).

- In order to run in dev mode, continue using `./scripts/serve-local.sh`. This just uses the react-app and comes with hot-reloading.
- In order to build the the static site, use the command `./scripts/build.sh`. This dumps the static files into the `/dist/` directory and maps them to your local root so you can preview how the static site works from that folder with a local server program like [`http-server`](https://www.npmjs.com/package/http-server).

Overall here are some things to note:
- `static.config.js` is an important file. In here, we make GraphQL queries to Joplin and pass them to each Page in the array that `getRoutes` returns.
- For each Page Component, the component gets exported using a HOC which returns the component with the data from the static.config passed in as props. ex: `getRouteProps(Department)` [more info](https://github.com/nozzle/react-static#withroutepropscomponent)

TODOS:
- [ ] Fix i18n. Currently not using the LanguageWrapper or i18n Select Language Banner. This fix to this may involve mapping over all the routes for each language code prefix in the static.config file.
- [ ] Fix Editor Preview when in Joplin. The way data is loaded has changed so we need to refactor so that the Live Previews work as they did before. This may involve wrapping specific page components in a parent component and serving them at a different preview specific route. 🤷‍♂️ 
- [x] Updates to script file for thinks like [this](https://github.com/cityofaustin/janis/blame/react_static/static.config.js#L11)
- [ ] Update deploy script and figure out a strategy to deploy static files. cc/ @benweatherman 
- [ ] Clean up fetchData file. Perhaps move queries out of static.config and into this one helper file.
- [x] ~Recollect isn't working, but it also isn't working in another PR preview deploy or in master. This might have to do with a separate commit but it still needs to get fixed I suppose.~


scripts/serve-local.sh
```diff
@@ -15,7 +15,10 @@ docker run \
     --volume "$PWD/src:/app/src" \
     --volume "$PWD/public:/app/public" \
     --volume "$PWD/package.json:/app/package.json" \
+    --volume "$PWD/static.config.js:/app/static.config.js" \
+    --volume "$PWD/.babelrc:/app/.babelrc" \
     --volume "$PWD/yarn.lock:/app/yarn.lock" \
     --env "REACT_APP_CMS_ENDPOINT=http://localhost:8000/api" \
+    --env "API_URL=$(ifconfig en0 | awk '$1 == "inet" {print $2}')" \
     --env "REACT_APP_CMS_ASSETS=http://localhost:8000/media" \
```
benweatherman: Do we need this URL anymore? I don't think `localhost` will work inside the container but might be fine for loading images.

src/index.js
```diff
@@ -1,7 +1,32 @@
...
+        <Comp />
+      </AppContainer>,
+      document.getElementById('root'),
+    )
+  }
+
+  // Render!
+  render(App)
+
+  // Hot Module Replacement
```
benweatherman: 🔥

src/js/helpers/fetchData.js
```diff
@@ -0,0 +1,50 @@
...
+
+// queries
+import allServicePagesQuery from 'js/queries/allServicePagesQuery';
+
+// clean data helpers
+import { cleanServiceLinks } from 'js/helpers/cleanData';
+
+export const fetchServices = (lang) => {
+  request(
+    `${process.env.REACT_APP_CMS_ENDPOINT}/graphql/`,
```
benweatherman: Should this be `API_URL`?

scripts/build.sh
```diff
@@ -0,0 +1,24 @@
...
+    --name janis \
+    --tty --interactive \
+    --volume "$PWD/dist:/app/_dist" \
+    --volume "$PWD/src:/app/src" \
+    --volume "$PWD/public:/app/public" \
+    --volume "$PWD/package.json:/app/package.json" \
+    --volume "$PWD/static.config.js:/app/static.config.js" \
+    --volume "$PWD/.babelrc:/app/.babelrc" \
+    --volume "$PWD/yarn.lock:/app/yarn.lock" \
+    --env "REACT_APP_CMS_ENDPOINT=http://localhost:8000/api" \
```
benweatherman: Do we need the `REACT_APP_*` vars?

src/js/helpers/fetchData.js
```diff
@@ -0,0 +1,50 @@
...
+      // this.setState({ data: data });
+      return data;
+    }
+  }
+
+  axios
+    .create({
+      headers: { 'Accept-Language': lang }
+    })
+    .post(`${process.env.REACT_APP_CMS_ENDPOINT}/graphql/`, {
```
benweatherman: Change to `API_URL`?

mateoclarke:
@benweatherman, on the Joplin side, I had to add a new item to the `ALLOWED_HOSTS` list when I would see an error like this.

screen shot 2018-01-30 at 1 36 13 pm

Wasn't sure if I should commit IPs like this to Joplin or just to keep them locally without committing.

benweatherman:
@mateoclarke sorry about that, lemme take a look at a more permanent fix on the joplin side

package.json
```diff
@@ -1,36 +1,42 @@
...
   "name": "janis",
-  "version": "0.1.0",
+  "version": "1.0.1",
   "description": "Public facing frontend public side of austin.gov using React components.",
-  "private": true,
+  "main": "index.js",
+  "license": "MIT",
   "dependencies": {
-    "axios": "^0.17.1",
+    "axios": "^0.16.2",
```
ifsimicoded: Any reason for the version rollback?

src/js/modules/Recollect.js
```diff
@@ -4,9 +4,7 @@ import Script from 'react-load-script';
 class Recollect extends Component {
 
   handleScriptLoad() {
-
     // Recollect is a third party script that isn't an import-able node module
-    // rCw is the required container id to ensure styles are shown correctly
```
ifsimicoded: minor, probably want to keep this comment about rCw

src/js/page_sections/Navmenu.js
```diff
@@ -106,20 +106,20 @@ class Navmenu extends Component {
...
               <li key={parentLink.id} onClick={this.props.toggleMenu}>
-                <Link to={getPathWithLangCode(`/topic/${parentLink.id}`)}
+                <Link to={`/topics/${parentLink.id}`}
                   className={this.getParentMenuItemClassName(parentLink)}
                 >
                   { parentLink.text }
                 </Link>
 
-                { serviceLinks.length && (
+                { !!serviceLinks && (
```
ifsimicoded: if serviceLinks is a zero length array, this will evaluate to true. I think we need to check length specifically.

src/js/pages/404.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+//
+
+export default () => (
+  <div>
+    <h1>404 - Oh no's! We couldn't find that page :(</h1>
```
ifsimicoded: WAT NO GIF?

scripts/build.sh
```diff
@@ -0,0 +1,23 @@
...
+    --name janis \
+    --tty --interactive \
+    --volume "$PWD/dist:/app/_dist" \
+    --volume "$PWD/src:/app/src" \
+    --volume "$PWD/public:/app/public" \
+    --volume "$PWD/package.json:/app/package.json" \
+    --volume "$PWD/static.config.js:/app/static.config.js" \
+    --volume "$PWD/.babelrc:/app/.babelrc" \
+    --volume "$PWD/yarn.lock:/app/yarn.lock" \
+    --env "CMS_API=http://$(ifconfig en0 | awk '$1 == "inet" {print $2}'):8000/api/graphql/" \
```
benweatherman: You should run the command for getting IP just once

Untested but something like
```bash
HOST_IP=$(ifconfig en0 | awk '$1 == "inet" {print $2}')
docker run \
...
  --env "CMS_API=http://$HOST_IP:8000/api/graphql/" \
  --env "CMS_MEDIA=http://$HOST_IP:8000/media/" \
```

scripts/serve-local.sh
```diff
@@ -15,7 +15,9 @@ docker run \
     --volume "$PWD/src:/app/src" \
     --volume "$PWD/public:/app/public" \
     --volume "$PWD/package.json:/app/package.json" \
+    --volume "$PWD/static.config.js:/app/static.config.js" \
+    --volume "$PWD/.babelrc:/app/.babelrc" \
     --volume "$PWD/yarn.lock:/app/yarn.lock" \
-    --env "REACT_APP_CMS_ENDPOINT=http://localhost:8000/api" \
-    --env "REACT_APP_CMS_ASSETS=http://localhost:8000/media" \
+    --env "CMS_API=http://$(ifconfig en0 | awk '$1 == "inet" {print $2}'):8000/api/graphql/" \
```
benweatherman: Same comment as `build.sh`

package.json
mateoclarke: nope, good catch. ty!

src/js/page_sections/Navmenu.js
mateoclarke: I was seeing a zero length array return `0` onto the DOM. I saw some twitter thread talking about this quirk in rendering elements with a map function but now I can't find it. That thread suggested using double bang so that's where I got the inspiration. I promise this fixed the bug I was seeing. But I can roll it back since I can't defend my actions with any evidence. Either way, if the navmenu doesn't return any list items, we have bigger problems.

mateoclarke:
Thanks everyone for the detailed review! @briaguya we started using axios because at first we were fetching data from the REST API and axios is chill. Then in the process of implementing React Static, I discovered that graphcool open-sourced a light weight query library for GraphQL so I decided to give it a shot and it helped me get to something working pretty quick. I replaced it for axios in the Navmenu. And the idea of refactoring the `fetchData` file (which still has some axios) might be worth punting if you want to go ahead and move towards using Apollo instead.

PR #120, opened 2018-01-29
Use new heroku build system

benweatherman:


benweatherman:
Merging so everyone can partake in the goodness

Issue #123, opened 2018-02-01
Update scripts/deploy.sh for deploying static files to CDN


mateoclarke:
Update deploy script and figure out a strategy to deploy static files. 

@benweatherman mentioned this strategy with Docker:
https://blog.onebar.io/reducing-docker-image-size-using-multi-stage-build-4ec8ee111aae

benweatherman:
This is done though now running into cityofaustin/janis#136

PR #127, opened 2018-02-02
Initial version of deploying the statically built site

benweatherman:
Now we'll default to a multi-stage build:
1. Build the static files
2. Copy static files from step 1 and serve via nginx

I've created a separate `Dockerfile.build` file, though I'm not sure if that'll be the right way to separate developing (with hot reloads, etc) and deployment. This is my first attempt and we'll iterate to get the right dev experience.

Implements cityofaustin/janis#123

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "start-js": "react-static start",
     "start": "npm-run-all -p watch-css start-js",
     "build-js": "react-static build",
-    "build": "npm-run-all build-css build-js && cp -r dist/* _dist/",
+    "build": "npm-run-all build-css build-js",
     "build-css": "node-sass-chokidar --include-path ./src --include-path ./node_modules src/ -o src/",
     "watch-css": "npm run build-css && node-sass-chokidar --include-path ./src --include-path ./node_modules src/ -o src/ --watch --recursive"
   }
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,0 +1,30 @@
+FROM node AS builder
+
+RUN mkdir /app
+WORKDIR /app
+
+ENV NODE_PATH=src
+
+COPY yarn.lock /app/yarn.lock
+COPY package.json /app/package.json
+RUN yarn
+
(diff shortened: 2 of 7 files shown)


benweatherman:
This is working when deployed, which is a big improvement over the current master branch 😛  Still not sure about keeping separate `Dockerfile`s and how this will work with local development. Let's chat about it.

package.json
```diff
@@ -35,7 +35,7 @@
     "start-js": "react-static start",
     "start": "npm-run-all -p watch-css start-js",
     "build-js": "react-static build",
-    "build": "npm-run-all build-css build-js && cp -r dist/* _dist/",
```
mateoclarke: will we still be able to test the static build locally without this?

benweatherman: Yes by running `./scripts/serve-build.sh`, though I'm not sure that's a great option because things won't get automatically rebuilt.

I made this change because I didn't want to copy from the nonstandard `_dist` in the Dockerfile https://github.com/cityofaustin/janis/pull/127/files#diff-d92c0cc6cc3b84dda3c92547bcf56981R28

I'll spend some more time to make sure `./scripts/serve-local.sh` works well for development.

benweatherman:
It's a good idea to describe how deploys work. I'll probably do that as a separate item somewhere since it's basically the same for joplin or other apps we'd host on heroku.

Regarding the other stuff

1. `app.json` is [required for pipelines to work](https://devcenter.heroku.com/articles/github-integration-review-apps#app-json). I did the bare minimum to get that going but there's all kinds of stuff it can do https://devcenter.heroku.com/articles/app-json-schema
2. It's a place for "stuff needed to run the container but isn't app code" but couldn't think of a better name than `deploy`. It'll be the same nginx config anywhere though, locally or deployed.

benweatherman:
The latest deployment is working and this shouldn't impact how local dev works (other than you can run `./scripts/serve-build.sh` to see how it'll work in nginx.

Issue #125, opened 2018-02-01
Refactor React components to stateless functions


mateoclarke:
We have a few components that aren't making use of React's component state or lifecycle methods. These components should be refactored to use the stateless functional pattern if they are purely for rendering DOM elements and not managing state. This will improve performance and clean up our code.

more info: https://tylermcginnis.com/functional-components-vs-stateless-functional-components-vs-stateless-components/

mateoclarke:
We've already been doing this as we touch the code, but just read another article that makes the case about why this is good:
http://reactingonrails.com/prefer-functional-components/

ifsimicoded:
Closing as all current components that should be stateless are.

PR #128, opened 2018-02-02
i18n 🌎 for React Static 

mateoclarke:
_Fixes #121_

Stuff happening in this PR:
- [x] update static.config to build pages with language code prefixes
- [x] update queries in static.config to reduce redundancy
- [x] refactor Page components to stateless functional pattern (#125)
- [x] reimplement `i18nBanner` & refactor `LanguageWrapper`
- [x] use request header to get API content by language


src/js/helpers/GoogleAnalyticsPageView.js
```diff
@@ -1,6 +1,8 @@
 import ReactGA from 'react-ga'
 
 const GoogleAnalyticsPageView = () => {
+  if (typeof document === 'undefined') return false;
```
mateoclarke: @ifsimicoded, This bombs in static build mode because window is undefined. But  loads progressively in the built pages. Is this a problem at all?

static.config.js
```diff
@@ -3,10 +3,10 @@ import { request } from 'graphql-request';
 
 // QUERIES
 import allServicePagesQuery from 'js/queries/allServicePagesQuery';
-import servicePageQuery from 'js/queries/servicePageQuery';
 import allTopicPagesQuery from 'js/queries/allTopicPagesQuery';
 import topicPageQuery from 'js/queries/topicPageQuery';
```
ifsimicoded: Minor comments for now as I wrap my head around this stuff.
Can we get rid of these single page queries (Topic, Service)?
Can we rename departmentPageQuery to AllDepartmentPagesQuery to match the others?
Do we use js/helpers/fetchData.js any longer too?

src/js/pages/Department.js
```diff
@@ -13,70 +13,60 @@ import FormFeedback from 'js/page_sections/FormFeedback';
 import Service311 from 'js/page_sections/Service311';
 import departmentPageQuery from 'js/queries/departmentPageQuery';
```
ifsimicoded: Still minor, can we kill these single page query imports (in the Department and Topic component files) since we're doing all page queries at the top level?

src/js/helpers/GoogleAnalyticsPageView.js
ifsimicoded: your change looks good. the React GA library uses `typeof window === 'undefined'`
https://github.com/react-ga/react-ga/blob/3d08ae2fbccf0209ae827a7f0128843301a3eff8/src/index.js#L81
Honestly, I don't know what's better -- the type check for undefined in a stateless functional component, or putting code you want to run on the client only in a ComponentDidMount function. The latter is how the Recollect script loads. I can update the Recollect component to be consistent.

src/App.js
```diff
@@ -1,55 +1,102 @@
...
+      return isChinese ? locale().toLowerCase() : twoLetterLangCode;
+    }
+  }
+
+  setLanguage = () => {
+    const cookieLanguage = Cookies.get('lang');
+    const browserLocale = this.parseBrowserLanguageCode();
+    let language = '';
+
+    const isLanguageCodeInPath = SUPPORTED_LANGUAGES
```
ifsimicoded: I don't think this chunk which pulls the language from the url path still works as is. I don't think we define this.props.urlPathLanguage any longer. This was something we were passing into LanguageWrapper from props.match.params.lang, correct?

ifsimicoded: This may be completely obvious, but I'm a total noob. I'm curious why you opted to pull the language specific logic out of a component and into App.js. I thought the abstraction kept the App.js code super clean.

src/js/helpers/GoogleAnalyticsPageView.js
mateoclarke: Seems like doing the document check is the best thing to do in a stateless function and putting this type of function in ComponentDidMount is the best thing to do if the component has state. Consistency is good, but I don't think we have enough examples to refactor towards "the best" pattern for these two different types of components.

static.config.js
mateoclarke: yes, yes, nope

src/App.js
mateoclarke: I think it still works, the language is stored in the component state and gets passed into the `IntlProvider` as the locale string. 

I just opted to pull it in because it was the most straightforward path to getting things to work. I'm open to moving it back into its own wrapper, but the App.js didn't seem overly complicated to me as it is.

mateoclarke: feb-06-2018 15-35-11

mateoclarke:
feb-06-2018 15-15-21

Going to call this a good stopping point and merge. One thing that isn't implemented here is rerouting links to the set language. @benweatherman and I talked about doing this at the ngnix level. So I'm going to punt on reimplementing the Links with the proper lang code prefix in the path.

More significantly, after digging deeper into React Static between this pull request and #119, I have some doubts about how well it would scale for us. For more info, #132.

src/App.js
ifsimicoded: Hey, I think I wasn't being clear. So if I navigate to a page with a language code in the url, It doesn't seem like the app recognizes this language code. So in the [video](https://drive.google.com/open?id=1TOxOlQ5Pkd1q4XSpCLSmYuHGiDDznFGz), I go to /es/ and you'll see that urlPathLanguage is undefined and english is the selected language in the menu and cookie instead of spanish.

mateoclarke: yup, you're right. working on fixing it now. Thanks for attaching the video

PR #134, opened 2018-02-08
Add special Hero image to Homepage

mateoclarke:
_Closes #103_

screen shot 2018-02-08 at 1 57 39 am
screen shot 2018-02-08 at 12 03 47 pm

- We'll probably want to pull this image in from Joplin in the future. How would we model Homepage content? 
- I'm doing a hacky thing in `Hero.js` with inline styles to get the overlay effect. Got it from [CSS Tricks](https://css-tricks.com/tinted-images-multiple-backgrounds/) but open to more elegant solutions.

TODO:
- [x] replace white search icon with black icon so it is visible in mobile views


mateoclarke:
Good call @benweatherman. Logged here: https://github.com/cityofaustin/joplin/issues/91

PR #138, opened 2018-02-13
Homepage Updates

mateoclarke:
_#33_

### DIFF
--- a/src/css/base/_variables_coa.scss
+++ b/src/css/base/_variables_coa.scss
@@ -25,3 +25,11 @@ $coa-textwrap-width: 545px;
   height: $height;
   justify-content: space-between;
 }
+
+// COA COLORS
+
+// BLUES
+$coa-color-blue: #005EFF;
+// PURPLES
+$coa-color-light-purple: #D7D2D9;
+$coa-color-dark-purple: #240B33;
--- a/src/js/modules/ExternalLink.js
+++ b/src/js/modules/ExternalLink.js
@@ -1,0 +1,11 @@
+import React from 'react';
+import ExternalLinkSVG from 'js/svg/ExternalLink'
+
+const ExternalLink = (props) => (
+  <a href={props.to} className="coa-ExternalLink">
+    {props.children}
+    <ExternalLinkSVG size="17" />
+  </a>
+);
+
+export default ExternalLink;
--- a/src/js/modules/ExternalLink.js
+++ b/src/js/modules/ExternalLink.js
@@ -1,10 +1,14 @@
 import React from 'react';
-import ExternalLinkSVG from 'js/svg/ExternalLink'
+import ExternalLinkSVG from 'js/svg/ExternalLink';
+
+const defaultIconSize = "17";
 
 const ExternalLink = (props) => (
-  <a href={props.to} className="coa-ExternalLink">
+  <a href={props.to} className="coa-ExternalLink"
+    target="_blank" rel="noopener noreferrer" aria-label="Opens in new window"
+  >
     {props.children}
-    <ExternalLinkSVG size="17" />
+    <ExternalLinkSVG size={props.iconSize || defaultIconSize} />
   </a>
 );
 
--- a/src/js/page_sections/Header.js
+++ b/src/js/page_sections/Header.js
@@ -33,7 +33,7 @@ class Header extends Component {
                 MENU
               </span>
               <span className="coa-text-spacer--vertical"></span>
-              <Link to={'/'}>AUSTIN.GOV</Link>
+              <Link to={'/'}>alpha.AUSTIN.GOV</Link>
             </div>
             <div className="col-xs-6 coa-Header__search">
               <Link to={'/search'}>
--- a/src/js/page_sections/Header.js
+++ b/src/js/page_sections/Header.js
@@ -33,10 +33,10 @@ class Header extends Component {
                 MENU
               </span>
               <span className="coa-text-spacer--vertical"></span>
-              <Link to={'/'}>alpha.AUSTIN.GOV</Link>
+              <Link to="/">ALPHA.AUSTIN.GOV</Link>
             </div>
             <div className="col-xs-6 coa-Header__search">
-              <Link to={'/search'}>
+              <Link to="/search">
                 Search <SearchSVG size="18"/>
               </Link>
             </div>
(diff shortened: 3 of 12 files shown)


src/js/modules/ExternalLink.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import ExternalLinkSVG from 'js/svg/ExternalLink'
+
+const ExternalLink = (props) => (
+  <a href={props.to} className="coa-ExternalLink">
```
ifsimicoded: should this link open in a new tab? the little icon makes me think it should, like the footer links, but I'm not sure.

src/js/page_sections/Header.js
```diff
@@ -33,7 +33,7 @@ class Header extends Component {
                 MENU
               </span>
               <span className="coa-text-spacer--vertical"></span>
-              <Link to={'/'}>AUSTIN.GOV</Link>
+              <Link to={'/'}>alpha.AUSTIN.GOV</Link>
```
ifsimicoded: Really minor, but any reason for the curly braces around the to property since this is static and not a variable? Same for the search link.

src/js/page_sections/SecondaryContentBanner.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const SecondaryContentBanner = (props) => (
+  <div className="coa-SecondaryContentBanner">
+    <div>
```
ifsimicoded: Any reason for the nested divs?

src/js/pages/Home.js
```diff
@@ -1,8 +1,17 @@
...
 import GlobalSearch from 'js/modules/GlobalSearch';
 import Hero from 'js/modules/Hero';
+import ExternalLink from 'js/modules/ExternalLink';
 import homepageImg from 'images/lady_bird_lake.jpg';
 
+import jsonFileData from '__tmpdata/services';
+const services311 = get(jsonFileData, "services311", null);
+
+
 class Home extends Component {
```
ifsimicoded: Our other pages are functional. It seems like the homepage doesn't require state, so could we make this consistent with other pages that are stateless components.

src/js/pages/Home.js
```diff
@@ -11,16 +20,22 @@ class Home extends Component {
...
         <Hero home={true} image={homepageImage}>
           <span className="coa-Hero__home-preheader">Welcome to</span>
           <h2 className="coa-Hero__home-header">Austin, TX</h2>
           <GlobalSearch placeholder="Search for services" />
         </Hero>
+        <SecondaryContentBanner content={bannerContent()} />
 	      <section className="wrapper coa-section">
-	        <h1> Home Page </h1>️
-	        <p>just a placeholder</p>
+          <Service311 services311={services311} />
```
ifsimicoded: The 311 module is already wrapped in a wrapper and section, so I don't think we need line 37 here and 39 to wrap it again.

src/css/base/_variables_coa.scss
```diff
@@ -25,3 +25,11 @@ $coa-textwrap-width: 545px;
   height: $height;
   justify-content: space-between;
 }
+
```
ifsimicoded: Not sure our colors are settled yet, but when they are, we should probably use these variables in _variables_uswds to overwrite primary, primary-alt and secondary colors.

ifsimicoded:
giphy 11

src/js/modules/ExternalLink.js
mateoclarke: yup, great call.

src/js/page_sections/Header.js
mateoclarke: no reason

src/js/page_sections/SecondaryContentBanner.js
mateoclarke: no reason

src/js/pages/Home.js
mateoclarke: yup, I was working on this when you jumped in to review. great minds...

src/css/base/_variables_coa.scss
mateoclarke: yeah agreed. I'm also thinking we might want to remove the USWDS dependency in a future refactor since there isn't a lot of championing of it within our team. I could be way off on this.

src/js/pages/Home.js
```diff
@@ -1,31 +1,40 @@
...
+        <GlobalSearch placeholder="Search for services" />
+      </Hero>
+      <SecondaryContentBanner content={bannerContent()} />
+      <Service311 services311={services311} />
+    </div>
+  );
 }
 
-export default Home;
+export default getRouteProps(Home);
```
ifsimicoded: Hey I think I read that getRouteProps was being deprecated for withRouteData https://github.com/nozzle/react-static/blob/0c0e8139acf045e309bba3a9ab9b91c295f437be/CHANGELOG.md
That said, I'm not 100% on why we use this -- mostly because I don't know 100% clearly how all this works. I thought this was used to pass the data we fetch from Joplin into our components -- but I don't see joplin data here. is this to future proof it?

static.config.js
```diff
@@ -118,27 +118,42 @@ export default {
       {
         path: '/',
         component: 'src/js/pages/Home',
+        getProps: () => ({
+          allServicePages: serviceQueries['en'],
+        }),
```
ifsimicoded: This is my own confusion about how react static works, but why are we running serviceQueries for the homepage? I'll be back Monday, and maybe we can pair/talk so I have a better understanding of what's going on. Thanks for being patient, cause I know you've explained this to me before.

src/js/pages/Home.js
```diff
@@ -1,31 +1,40 @@
...
-          <h2 className="coa-Hero__home-header">Austin, TX</h2>
-          <GlobalSearch placeholder="Search for services" />
-        </Hero>
-	      <section className="wrapper coa-section">
-	        <h1> Home Page </h1>️
-	        <p>just a placeholder</p>
-	      </section>
-			</div>
-    );
+const Home = ({ allServicePages }) => {
```
ifsimicoded: allServicePages --> is this for the future? Why do we need service page data here?

mateoclarke:
@ifsimicoded We'll need services on the homepage for the update mocks. The homepage has "tiles" for 4 services. I was starting to work on that when @benweatherman asked me to go ahead and merge the PR as it was because PIO was causing a fire drill to get the "alpha.austin.gov is a work in progress, blah, blah blah..." content live. Blame it the suits.

Issue #18, opened 2017-11-21
Feedback Form


benweatherman:
- [x] UI Views for each state
- [x] Client form validation
- [x] Integration with Microservice Form Dispatcher
- [x] Workflow Analytics [[see below]](#issuecomment-350322802)
- [ ] figure out how we want to log email addresses -- currently shows in issue, which isn't very private.
- [ ] determine what we want to do in production vs development mode.

### Mockup
![screen shot 2018-01-11 at 9 20 31 am](https://user-images.githubusercontent.com/20911665/34996233-9c088bd4-fa9e-11e7-83c3-661ec9f78264.png)

**language translation only shows up as an option on a translated paged (re: see original mock options)**

### Step-by-step

1. User opts to submit feedback on page
2. Following options pop up:
  - Information on this page
  - Language translation (only shows up if page is translated from English to another language)
  - Technical issue
  - Something else
3. User chooses option which prompts a specific ask on the feedback form' (screen 4 above) =

- [ ] If selected "information on this page" = How can we improve this page?

- [ ] If selected "technical issue" = What technical issue are you having?

- [ ] If page is translated to a non-English language, and selected "language translation" = How can this page translation be improved?

- [ ] If selected "something else" = How can we help you?

  - Field appears with notification that feedback is public
  - Description field - rich text - required 
4. User sends feedback, form captures
  - Type of feedback
  - Feedback itself
  - URL, Device, Browser, OS

### More Info

Feedback form only shows up on Service Pages or other Page Templates–not topic or theme pages

Check out [further comments & context](https://docs.google.com/document/d/1t1FwkMRrGwRyh3NbJwqaSDHLTnLU2DkplPq_aRRcJj8/edit#)

cthibaultatx:
![screen shot 2017-11-28 at 1 00 46 pm](https://user-images.githubusercontent.com/20911665/33338791-bc604030-d43c-11e7-8888-46ed7ce54daa.png)

mattlangan:
I believe the plan is for this form to submit to the [Form Dispatcher](https://github.com/cityofaustin/forms-api-dispatch), which will then kick the submission off to the [GitHub API Issue Creator](https://github.com/cityofaustin/api-github-issue) service. Initial documentation is available in the repositories and includes required and optional params and server responses.

amenity:
It looks like this issue misses another requested function; returning the issue URL to user. 
 
Returns link to issue in website submission confirmation modal: 
[...]

PR #151, opened 2018-02-14
Feedback Posts to Janis' repo Issues

ifsimicoded:
completes an additional step of #18.
- [ ] figure out how we want to log email addresses -- currently shows in issue, which isn't very private.
- [ ] determine what we want to do in production vs development mode.


scripts/serve-local.sh
```diff
@@ -19,6 +19,7 @@ docker run \
     --volume "$PWD/static.config.js:/app/static.config.js" \
     --volume "$PWD/.babelrc:/app/.babelrc" \
     --volume "$PWD/yarn.lock:/app/yarn.lock" \
+    --env "FEEDBACK_API=https://coa-test-form-api.herokuapp.com/process/" \
```
mateoclarke: Should this be set as an environment variable if it is the same for all the environments?

src/js/page_sections/FormFeedback.js
```diff
@@ -85,10 +88,20 @@ class FormFeedback extends Component {
...
-    // after successful form submit
-    this.setState({
-      stage: 0
+
+    postFeedback({
+      title: this.state.values['site-feedback-options'],
+      description: this.state.values['site-feedback-textarea'],
+      email: this.state.values['site-feedback-email']
+    })
+    .then(({data: data}) => {
```
mateoclarke: could this line just be `.then(({ data }) => {`

src/js/page_sections/FormFeedback.js
```diff
@@ -269,7 +282,7 @@ class FormFeedback extends Component {
           <div className="coa-overlay">
             <div className="coa-overlay__content">
               <h4>Thank you for sharing your feedback!</h4>
-              <p>You can see your feedback in our <a href="#">austin.gov feedback tracker</a></p>
+              <p>You can see your feedback in our <a href={this.state.successUrl} target="_blank" rel="noopener noreferrer" aria-label="Opens in new window">austin.gov feedback tracker</a></p>
```
mateoclarke: 👍

mateoclarke:
Full story form abandonment looks really cool.

ifsimicoded:
Hey y'all, sorry I forgot to create a new branch, so I updated this PR with event tracking. If y'all could review.
https://github.com/cityofaustin/janis/pull/151/commits/460cba7201437e903fe90b62912d0768f084d619

src/App.js
```diff
@@ -1,21 +1,21 @@
 import React, { Component } from 'react'
 import { Router, Route } from 'react-static'
 import ReactGA from 'react-ga'
-import GoogleAnalyticsPageView from 'js/helpers/GoogleAnalyticsPageView'
+import { logPageView } from 'js/helpers/googleAnalytics'
```
mateoclarke: I like this.

scripts/serve-local.sh
ifsimicoded: I'm thinking we'll have different endpoints in different environments (with prod having more rigorous validation). We probably don't want issues being created in dev/staging as is currently the case, and I haven't thought through what the long term plan is, but this leaves it open. Also cool with changing -- lemme know what you're thinking.

src/js/page_sections/FormFeedback.js
ifsimicoded: yep!

ifsimicoded:
@mateoclarke if you have implementation changes, I'm down to talk/pair.

PR #158, opened 2018-02-24
Homepage Services List

### LINKED ISSUE #33: Implement homepage design

mateoclarke:
Closes #33 

- refactor List Links to use "Tile" as mocked up in Sketch
- refactor some stateful components to stateless functions
- clean up Secondary content banner on home page.
- made some guesses about hover states that should be reviewed with design.

screen shot 2018-02-24 at 1 28 27 am

screen shot 2018-02-24 at 1 27 59 am


mateoclarke:
Yeah, same error as before. I checked the deploy once it spun up but it was late so I gave up for the night.

mateoclarke:
Home is fixed, but we'll need to come back and fix the index pages for `/services` and `/topics`. But since those aren't linked anywhere, I'm not worried about it yet.

ifsimicoded:
The failure to render on initial page load, I think has something to do with content on the Recycle Christmas Tree Page because that page also seems to be experiencing the same failure to render on initial load `Uncaught SyntaxError: Invalid or unexpected token`. Just FYI. I think if we remove/hide that page from the CMS, it would be a temporary band-aid for all service page queries.

static.config.js
```diff
@@ -115,43 +120,43 @@ export default {
...
     return [
+      ...allPages(),
       {
         path: '/',
         component: 'src/js/pages/Home',
-        // getProps: () => ({
-        //   allServicePages: serviceQueries['en'],
-        // }),
+        getData: async () => ({
+          allServicePages: topServices,
```
ifsimicoded: We'll need to add topServices query (or kill the x-mas tree page) to homepage requests in all languages. Currently the homepage in spanish, vietnamese, etc experience the same initial load failure.

static.config.js
```diff
@@ -33,12 +34,16 @@ export default {
     const { allServicePages } = await enGraphQLClient.request(allServicePagesQuery)
     const { allTopics } = await enGraphQLClient.request(allTopicPagesQuery)
     const { allDepartments } = await enGraphQLClient.request(allDepartmentPagesQuery)
+    const { allServicePages: topServices } = await enGraphQLClient.request(topServicesQuery)
```
ifsimicoded: super minor, but could be update the allServicePages var naming everywhere related to the homepage to be more descriptive to the returned data if we decide to keep this query. I got confused about why we were pulling all service page data on the homepage.

src/js/svg/ChevronCircleRight.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+function ChevronCircleRight(props) {
```
ifsimicoded: SVG 👍

src/js/queries/topServicesQuery.js
```diff
@@ -0,0 +1,19 @@
+const topServicesQuery = `
+  query topServices {
+    allServicePages(first: 4) {
```
ifsimicoded: just in line with the naming comment below -- I think allServicePages is sorta confusing.

src/js/modules/Tile.js
```diff
@@ -0,0 +1,22 @@
...
+
+const Tile = ({ url, text, style, tag }) => (
+  <Link className="coa-Tile" to={url}>
+    <span className="coa-Tile__tag">{tag}</span>
+    <p className="coa-Tile__text">{text}</p>
+    <ChevronCircleRightSVG size="22"/>
+  </Link>
+);
+
+Tile.propTypes = {
```
ifsimicoded: A+ on using prop types ✅  But, McKinley asked a good question that lead us to find that React-Static doesn't acknowledge proptypes -- or if it does, doesn't throw any warnings when proptypes are not used. Not something to worry about now, but might be something we think about when we come up with a build plan that includes testing.

src/css/page_sections/_SecondaryContentBanner.scss
```diff
@@ -4,8 +4,10 @@
   background-color: $coa-color-light-purple;
   color: $coa-color-dark-purple;
   p {
+    max-width: 100% !important; // TODO: talk to simi about coa-main
```
ifsimicoded: OH NO! let's talk about why you need this.
The 545px max-width was set for pages with content from the CMS -- so the content/text wouldn't extend beyond a readable width. 

Could this p be a div instead? Then we wouldn't need the max-width declaration. Or maybe the mocks have changed enough that we should probably just toss the existing css, and start over instead of trying to fight with it with !important.

src/css/modules/_SectionTitle.scss
```diff
@@ -0,0 +1,16 @@
+.coa_SectionTitle {
+  margin: $coa-space-level7 auto $coa-space-level6;
+  text-transform: uppercase;
+  text-align: center;
+  font-size: $h2-font-size;
+  font-weight: bold;
+  max-width: 100% !important; // needed to override coa0-main textwrap. TODO: talk to simi
```
ifsimicoded: Oh, this is similar to the comment below. @cthibaultatx , initially had limitations on text width to keep it readable on large screens. I'm not sure if she intended to stick with the limitation, but with text centered instead. Just a place to get clarification. Also, like below, we may wanna toss existing css so we don't fight with it. Seems like things have changed enough to do so.

static.config.js
mateoclarke: Thanks for catching this. I got excited that changing the query fixed things but forgot to take it across the finish line for the other languages.

static.config.js
mateoclarke: another good call.

src/js/queries/topServicesQuery.js
mateoclarke: I don't think its as simple as renaming without changing the joplin code. 
```
{
  "errors": [
    {
      "locations": [
        {
          "column": 5,
          "line": 2
        }
      ],
      "message": "Cannot query field \"topServices\" on type \"Query\". Did you mean \"allServicePages\"?"
    }
  ]
}
```
It makes ok sense to me that the topServices would be a subset (the first 4) of all service pages. I think it might make sense to have a topServices query in joplin in the future, but maybe it would be best as an editable within the CMS featuredServices? idk.

src/js/modules/Tile.js
mateoclarke: good catch McKinley. but, damn that's frustrating though. I'm over React-Static. I've also noticed that it suppresses the basic linting we used to get with create-react-app.

src/css/page_sections/_SecondaryContentBanner.scss
mateoclarke: Needed to do this for the banner on the homepage...

screen shot 2018-02-26 at 10 28 09 pm

If it was intended for CMS content pages, could we apply those rules with a new class like `.coa-content` instead of applying it across the board? down to talk about this more IRL.

mateoclarke:
Yup, I'm down for opening an issue about the weird issue with the christmas page content. But also I'm ready to start thinking about what Server Side Rendered or Universal React framework we might want to port over to instead of digging deeper into React-Static. That feels like the discussion we need to have soonish.

src/js/queries/topServicesQuery.js
mateoclarke: @benweatherman, any thoughts about what this query might be?

src/js/modules/Tile.js
mateoclarke: also down to add this as an issue to the backlog.

static.config.js
mateoclarke: fixed this with my pr feedback commit.

PR #159, opened 2018-02-28
Language Updates

### LINKED ISSUE #89: Create a Wrapper Component for <Link> that accounts for language codes
See discussion: https://github.com/cityofaustin/janis/pull/86#discussion_r161315298

[basename](https://reacttraining.com/react-router/web/api/BrowserRouter/basename-string) in Browser Router  would automatically do this for us.

ifsimicoded:
- Manage language setting from path/match props only in language wrapper
- Fixes showing correct content if language is set via cookie instead of route
- Add language code to all links ( Closes #89 )
- Clean up link active class management in navmenu and language select banner
- Clean up props management


src/js/modules/I18nLink.js
```diff
@@ -0,0 +1,11 @@
...
+import { Link, withRouteData } from 'react-static'
+import { i18nalizeLinkTo } from 'js/constants/languages'
+
+const I18nLink = (props) => {
+  const {langCode, to, ...rest} = props;
+  //TO DO -- only pass Link props, not all remaining ...rest
+  return <Link to={i18nalizeLinkTo({to, langCode})} {...rest} />
+}
+
+export default withRouteData(I18nLink);
```
mateoclarke: Why do we need to wrap `I18nLink` with the Higher Order Component for RouteData?

src/js/page_sections/LanguageSelectBanner.js
```diff
@@ -80,32 +40,32 @@ class LanguageSelectBanner extends Component {
...
                 }
-                <div className={this.getMenuClassName()}>
+                <div className={`coa-LanguageSelectBanner__dropdown-menu ${this.state.isOpen ? 'coa-LanguageSelectBanner__dropdown-menu--open' : ''}`}>
                   <div className="wrapper">
                     <h4 className="coa-LanguageSelectBanner__menu-header">
                       Translated by City of Austin
                     </h4>
                     <ul className="coa-LanguageSelectBanner__language-list">
-                      { SUPPORTED_LANGUAGES.map((language, i) => {
+                      { SUPPORTED_LANGUAGES.map(({title, abbr, code}, i) => {
```
mateoclarke: 👍

src/js/page_sections/Navmenu.js
```diff
@@ -84,17 +51,19 @@ class Navmenu extends Component {
 
     return parentLinks.length && (
       <div className="usa-grid-full">
-        <nav className={this.menuClassName()}>
+        <nav className={`coa-Navmenu ${this.props.isOpen ? 'coa-Navmenu--open' : ''}`}>
```
mateoclarke: Seems like you prefer this inline logic rather than calling a function that returns the proper string. Can you explain why this wasn't working well for you? For me, the code reads easier when the markup is clean and the classname comes back from a function.

static.config.js
```diff
@@ -77,13 +77,15 @@ export default {
         path: '/services',
         component: 'src/js/pages/Services',
         getData: async () => ({
-          allServicePages: serviceQueries[langCode],
+          allServices: serviceQueries[langCode],
+          langCode: langCode,
         }),
-        children: serviceQueries[langCode].edges.map(service => ({
-          path: `/${service.node.slug}`,
+        children: serviceQueries[langCode].edges.map(({node: service}) => ({
```
mateoclarke: 👍 x 3

static.config.js
```diff
@@ -106,18 +110,20 @@ export default {
...
-            department: department.node,
+            department: department,
+            langCode: langCode,
           })
         }))
       },
       {
         path: '/search',
-        component: 'src/js/pages/Search',
+        component: 'src/js/pages/Search', //TODO: update 404 page to be conscious of all languages
```
mateoclarke: I think in this comment you meant `//TODO: update Search page to be conscious of all languages`

Which also makes me think about something for the first time. How do we support multi-language search. cc/ @benweatherman

static.config.js
```diff
@@ -160,12 +169,13 @@ export default {
...
         children: allPages('ar'),
         getData: async () => ({
+          langCode: "ar",
           topServices: topServices_ar,
         }),
       },
       {
         is404: true,
-        component: 'src/js/pages/404',
+        component: 'src/js/pages/404', //TODO: update 404 page to be conscious of all languages
```
mateoclarke: Good catch. We should get this on the content team's radar as an issue. We'll need the language a user should see when they 404. Also could be an opportunity for them to have some fun: https://www.creativebloq.com/web-design/best-404-pages-812505

Then, we need to get that language translated.

mateoclarke:
One thing I'm thinking about is, will it be easy for us to remember to use the `i18nLink` components? Would it be easier if they were named `Link` and imported from a local module instead of a library like React-Static? But then what happens if we need both our custom Link and a library Link in the same component, we'd have a namespace clash. nvm...

src/js/page_sections/Navmenu.js
ifsimicoded: Totes true, with it being cleaner in a method.
This was phase one, before I import classnames, so all this class stuff can be inline and even cleaner -- without ternaries inline.

static.config.js
ifsimicoded: yep. and YIKES.

src/App.js
```diff
@@ -13,16 +13,7 @@ class App extends Component {
...
-            render={(props) => {
-              logPageView();
-              return (
-                <LanguageWrapper {...props}
-                  urlPathLanguage={props.match.params.lang}
-                />
-              )
-            }}
-          />
+          <Route path="(/)?:lang([a-z]{2})?/:path*" component={LanguageWrapper} />
```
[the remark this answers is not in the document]
ifsimicoded: Can do -- this basically matches any two lowercase characters and throws them in the lang match prop if it exists. and everything else gets thrown in the path match prop. This way we can have the url path without the language handy too for the language select banner. This was simpler than splitting with js and pulling it out manually.

src/css/page_sections/_LanguageSelectBanner.scss
```diff
@@ -36,39 +36,28 @@
 }
```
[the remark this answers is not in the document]
ifsimicoded: OH. I think we'll probably end up doing a pretty big css refactor for the latest mocks, but I can update this for sure.

src/js/components/LanguageWrapper.js
```diff
@@ -9,93 +8,99 @@ import Cookies from 'js-cookie'
...
-      const twoLetterLangCode = locale().split('-')[0].toLowerCase();
-      // But we do want to support two types of Chinese (zh-tw & zh-cn)
-      const isChinese = twoLetterLangCode === 'zh';
-      return isChinese ? locale().toLowerCase() : twoLetterLangCode;
-    }
+  getInitialLangState() {
+    const getLang = [ this.getLangFromProps, this.getLangFromCookie, this.getLangFromLocale, ()=>DEFAULT_LANG ];
+    const that = this;
+
+    return getLang.reduce((lang, fn) => {
```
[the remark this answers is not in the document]
ifsimicoded: can you tell I literally *JUST* ready a medium article on declarative programming?????

src/js/components/LanguageWrapper.js
```diff
@@ -9,93 +8,99 @@ import Cookies from 'js-cookie'
...
   }
 
-  handleLanguageUpdate = (newLang) => {
-    Cookies.set('lang', newLang, { expires: this.daysUntilCookieExpires })
-    this.setState({ lang: newLang })
+  getLangFromLocale() {
+    if (typeof document === 'undefined') return null;
+
+    const lang = this.getSupportedLang(locale().split('-')[0].toLowerCase());
+    // MVP does not need to include support for two types of Chinese (zh-tw & zh-cn)
```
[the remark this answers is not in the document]
ifsimicoded: 👍

src/js/modules/I18nLink.js
ifsimicoded: Because we need the langCode, which the withRouteData HOC passes in as props.

static.config.js
benweatherman: Putting all the languages on a single 404 page could be fun and seems like would make things easier

mateoclarke: #resistance 

image

src/js/components/LanguageWrapper.js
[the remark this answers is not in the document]
mateoclarke: 🔥 so hawt right now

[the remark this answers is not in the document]
mateoclarke: and why are you holding back on us?! share that declarative shiz article in #dev plz.

Issue #124, opened 2018-02-01
Update Global Navigation


cthibaultatx:
- [ ] add theme level navigation

- [ ] show theme and topic level nav in mobile menu

- [ ] list of themes and topics (being finalized Friday 2/2): https://airtable.com/shr3RatfCvSZmJhqw

copy/pasting checklist items from #33 that belong here:

- [ ] ...refine mobile and tablet menu and develop new large/desktop menu

- [ ] Themes and topics are listed here: [Future Navigation airtable](https://airtable.com/shr3RatfCvSZmJhqw) Only build in topics we already have services/content for ('Recycling, Trash, and Compost' under 'Housing & Utilities'). Leave off other topics and instead, when a user hovers over a theme with no topics- they should read: (waiting on copy from content team @courtneyjacinic )

- [ ] add airport icon to tablet and desktop header, and in mobile menu- links to http://austintexas.gov/department/airport

- [ ] add 311 link to header to tablet and desktop, and in mobile menu- links to http://311.austintexas.gov/reports/list_services

- [ ] add 'alpha' to 'austin.gov' = 'alpha.austin.gov'

![screen shot 2018-02-16 at 12 39 58 pm](https://user-images.githubusercontent.com/20911665/36400789-bef4b7e8-1598-11e8-8675-0867b6417827.png)
![screen shot 2018-02-16 at 12 40 16 pm](https://user-images.githubusercontent.com/20911665/36400790-bf03510e-1598-11e8-80d2-c5260be3d130.png)
![screen shot 2018-02-16 at 12 40 42 pm](https://user-images.githubusercontent.com/20911665/36400796-c8f94f56-1598-11e8-9193-1f00007ffc26.png)
![screen shot 2018-02-16 at 12 42 49 pm](https://user-images.githubusercontent.com/20911665/36400797-c9085f6e-1598-11e8-90dc-128229316d69.png)

#### Behaves like: https://www.mass.gov/

mateoclarke:
duplicate of https://github.com/cityofaustin/techstack/issues/106

PR #160, opened 2018-03-01
Updated Global Nav

mateoclarke:
#124 
cityofaustin/techstack#106

Adding in a utility from Bootstrap v4 to help us with responsive display properties:
https://getbootstrap.com/docs/4.0/utilities/display/

TODO:
- [x] Clarify what themes and topics should be visible. Items 2 & 3 in [this description](https://github.com/cityofaustin/techstack/issues/106) seem to conflict.
   - [x] update links
   - [x] make a 'coming soon' placeholder page
- [ ] Double check a11y tab through behavior. 
   - Check markup, add aria tags?
   - skip nav
- [ ] Styling for active page, (need to confirm this in the design)
- [ ] Fix positioning of far right nav list so it doesn't fall of the screen

mar-07-2018 02-43-16

mar-07-2018 02-03-09


mateoclarke:
There's still some work to do but this PR is getting big so I'll take on the rest of my todo list in a new branch so I can get y'alls review. 🙏

src/__tmpdata/navigation.json
```diff
@@ -0,0 +1,194 @@
+{
+  "themes": [
```
benweatherman: Would it make sense to structure this more like a graphql response? That might mean less changes once we put this in the CMS. The `allTopics` response is probably the best starting point for a structure like that.

```
{
  "data": {
    "allThemes": {
      "edges": [
        {
          "node": {
            "title": "womp womp",
            "slug": "womp-womp"
```

etc

src/js/page_sections/Navmenu.js
```diff
@@ -1,115 +1,164 @@
...
       this.focusOnClose();
     }
   }
 
-  getOverlayClassName = () => {
-    let className = `coa-Navmenu__overlay`;
-    if (this.props.isOpen) {
-      className = `${className} ${className}--open`;
+  toggleMobileSublist = (e, openSectionId) => {
+    if (window.innerWidth < this.DESKTOP_BREAKPOINT) {
```
benweatherman: Does something like this work for SSR? Do we care about that now?

mateoclarke: hmm, interesting point, but this function only executes on a click event, so I don't think we need to expect this to be node/server-safe.

And if it works for React-Static, then I don't see why it wouldn't work in SSR. And it currently works in the PR deploy. 

But good eye to think about using the `window` with care. I wasn't thinking about that when I wrote this.

PR #161, opened 2018-03-02
google analytics reimplementation

ifsimicoded:
- Make google analytics account number an environment variable
@benweatherman is there a better way to make/set environment variables? Defining things in 3 different places seems fragile.
- reimplement pageviews -- the last PR I removed it.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -15,9 +15,6 @@ RUN yarn
 COPY static.config.js /app/static.config.js
 COPY .babelrc /app/.babelrc
 
-ARG GOOGLE_ANALYTICS
-ENV GOOGLE_ANALYTICS=${GOOGLE_ANALYTICS:-UA-110716917-2}
-
 ARG FEEDBACK_API
 ENV FEEDBACK_API=${FEEDBACK_API:-https://coa-test-form-api.herokuapp.com/process/}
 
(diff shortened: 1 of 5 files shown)


Dockerfile.build
```diff
@@ -15,6 +15,9 @@ RUN yarn
 COPY static.config.js /app/static.config.js
 COPY .babelrc /app/.babelrc
 
+ARG GOOGLE_ANALYTICS
+ENV GOOGLE_ANALYTICS=${GOOGLE_ANALYTICS:-UA-110716917-2}
```
benweatherman: We don't need to define a default value here

Dockerfile.build
```diff
@@ -15,9 +15,6 @@ RUN yarn
 COPY static.config.js /app/static.config.js
 COPY .babelrc /app/.babelrc
 
-ARG GOOGLE_ANALYTICS
-ENV GOOGLE_ANALYTICS=${GOOGLE_ANALYTICS:-UA-110716917-2}
```
benweatherman: I just meant we'd be able to remove the default value here
```
ARG GOOGLE_ANALYTICS
ENV GOOGLE_ANALYTICS=${GOOGLE_ANALYTICS}
```

PR #163, opened 2018-03-05
More Language Code In Links

### LINKED ISSUE #162: Update Internationalized Link Components
Currently the I18nLink and I18nNavLink are receiving and passing through more props than a Link and NavLink component expect respectively. This is causing the extra props to be added as attributes on the final rendered anchor tags and some warnings to be thrown in console.

screen shot 2018-03-05 at 11 48 26 am

We need to update both I18nLink and I18nNavLink components to only pass through expected props. Consider using https://lodash.com/docs/4.17.5#pick or https://stackoverflow.com/questions/17781472/how-to-get-a-subset-of-a-javascript-objects-properties

ifsimicoded:
I simplified how language code was handled -- instead of passing language (and a bunch of unnecessary data) through react static RouteData props I'm utilizing context to pass ONLY langCode from parent (LanguageWrapper) to children (I18nLink and I18nNavLink components).

This will close #162 and also is more react-static neutral and hopefully usable when we toss that for another SSR library.

### DIFF
--- a/src/js/components/LanguageWrapper.js
+++ b/src/js/components/LanguageWrapper.js
@@ -3,6 +3,7 @@ import Routes from 'react-static-routes'
 import { IntlProvider } from 'react-intl'
 import locale from 'browser-locale'
 import Cookies from 'js-cookie'
+import PropTypes from 'prop-types'
 
 // page_sections
 import LanguageSelectBanner from "js/page_sections/LanguageSelectBanner"
@@ -17,6 +18,12 @@ class LanguageWrapper extends Component {
     this.state = {
       lang: this.getInitialLangState()
     }
+  }
+
+  getChildContext() {
+    return {
+      langCode: this.state.lang
+    };
   }
 
   getInitialLangState() {
@@ -98,6 +105,10 @@ class LanguageWrapper extends Component {
       </IntlProvider>
     );
   }
+}
+
+LanguageWrapper.childContextTypes = {
+  langCode: PropTypes.string
 }
 
 export default LanguageWrapper;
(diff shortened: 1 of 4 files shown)


src/js/components/LanguageWrapper.js
```diff
@@ -100,4 +107,8 @@ class LanguageWrapper extends Component {
   }
 }
 
+LanguageWrapper.childContextTypes = {
+  langCode: PropTypes.string
```
mateoclarke: 👏

src/js/components/LanguageWrapper.js
```diff
@@ -19,6 +20,12 @@ class LanguageWrapper extends Component {
     }
   }
 
+  getChildContext() {
```
mateoclarke: I didn't know much about this React method. Seems like a good use case for what we're trying to do here! I definitely like it better than the React-Static specific HOC pattern.

I found this article that also lists out a few things to be careful of with React context. Many just gotchas around debugging and (over)coupling together two components.
https://javascriptplayground.com/context-in-reactjs-applications/

If you have any good resources you read that gave you the idea to implement this pattern, please share, I'd love to learn more.

ifsimicoded: for sure, I've mostly been looking at [React docs](https://reactjs.org/docs/context.html) straight up and how Links and Routers work in react-router-v4 -- since Links were super nested all over and need to have Routers in scope, I was just looking at how that was handled.  Thanks for the info link -- I definitely agree that context can complicate things when overused, but here it seems like the right solution, since Links, and maybe other components, might need to be aware of langCode for translation.

PR #164, opened 2018-03-08
react-intl implementation

ifsimicoded:
Janis static content language translation

@benweatherman @mateoclarke -- this is a reminder for me to solicit your feedback on when in our build/serving process we should build the json that's needed for translated content (done via react-intl + babel plugin and intl.buildlangs.js). That probably sounds unclear and wacky. I'll explain in person.


.gitignore
```diff
@@ -21,4 +21,5 @@ npm-debug.log*
 yarn-debug.log*
```
ifsimicoded: just curious if we want to clear out the defaults in the .gitignore that aren't relevent to our use case

Dockerfile.build
```diff
@@ -27,6 +28,7 @@ ENV CMS_API=${CMS_API:-https://joplin.herokuapp.com/api/graphql/}
 ARG CMS_MEDIA
 ENV CMS_MEDIA=${CMS_MEDIA:-https://joplin.herokuapp.com/media/}
 
+COPY public /app/public
```
ifsimicoded: saw this was in Dockerfile but not in Dockerfile.build -- just I added it. Not sure it's necessary at this moment tho.

ifsimicoded:
Closing this in favor of a new PR that has a less wonky history. Rebasin'

PR #165, opened 2018-03-08
Sd followup reactintl

ifsimicoded:
Janis static content language translation
- setup infrastructure for translated static content
- setup infrastructure for managing translated content (babel + intl.buildlangs.js -- note, we have to manually run this script whenever we want to generate a new default.json file which is the base json file for our translated content. @benweatherman @mateoclarke let me know if you think manually running this sucks and if there's a way to integrate it into our build process)
- updated stateful components to stateless functions where possible
- added homepage content which should live in Joplin to react static config


.gitignore
```diff
@@ -21,4 +21,5 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
+# sass to css
```
ifsimicoded: just curious if we want to clear out the defaults in the .gitignore that aren't relevent to our use case

Dockerfile.build
```diff
@@ -27,6 +28,7 @@ ENV CMS_API=${CMS_API:-https://joplin.herokuapp.com/api/graphql/}
 ARG CMS_MEDIA
 ENV CMS_MEDIA=${CMS_MEDIA:-https://joplin.herokuapp.com/media/}
 
+COPY public /app/public
```
ifsimicoded: saw this was in Dockerfile but not in Dockerfile.build -- just I added it. Not sure it's necessary at this moment tho.

.gitignore
mateoclarke: seems good

Dockerfile.build
```diff
@@ -12,6 +12,7 @@ COPY yarn.lock /app/yarn.lock
 COPY package.json /app/package.json
 RUN yarn
 
+COPY intl.buildlangs.js /app/intl.buildlangs.js
```
mateoclarke: we'll need to add this to the Docker.storybooks too huh.

README.md
```diff
@@ -37,8 +20,23 @@ Add a package via yarn (https://yarnpkg.com/lang/en/docs/cli/add/)
...
+___
+
+## Updating translation export via yarn
+
+All static translations live in src/js/i18n/locales/ directory.
+These files are versioned and built from the auto-generated default.json file (within that directory) via [babel-plugin-react-intl](https://github.com/yahoo/babel-plugin-react-intl).
+To re-generate default.json to update the static content to be translated, the following command can be run:
+
+```
+docker exec --interactive --tty janis yarn run build-langs
```
mateoclarke: 🔥

Dockerfile.build
ifsimicoded: nah, the build langs is only to generate the base english json -- so it's only needed in a dev environment

PR #172, opened 2018-03-19
Header v2

mateoclarke:
_One part of https://github.com/cityofaustin/techstack/issues/180_

screen shot 2018-03-20 at 11 43 44 am

screen shot 2018-03-20 at 11 43 36 am


src/js/page_sections/Menu/Menu.js
```diff
@@ -8,6 +9,33 @@ import navigation from '__tmpdata/navigation';
 import CloseSVG from 'js/svg/Close';
 import citySealImg from 'images/coa_seal.png';
 
+const i18nMessages = defineMessages({
+  airport: {
```
ifsimicoded: Probably need a home translation too for HomeMobileListItem

src/js/page_sections/Menu/Menu.js
```diff
@@ -55,8 +84,8 @@ class Menu extends Component {
           </button>
           <ul className="coa-Menu__list">
             <HomeMobileListItem handleClick={this.props.toggleMenu} />
-            <AirportMobileListItem />
-            <ThreeOneOneMobileListItem />
+            <AirportMobileListItem {...this.props} />
```
ifsimicoded: can we pass just the intl provider instead of all props so it's more obvious what we actually need/use in the sub components.

mateoclarke:
Made an issue about Menu squishing. https://github.com/cityofaustin/janis/issues/176 

This Header PR is different from the Menu work I already merged.

PR #175, opened 2018-03-20
More Service Page

ifsimicoded:
Finishes out second half of Service Page

- [x] extra content / Markup from Joplin
- [x] Recollect Calendar
- [x] Static Map
- [x] Contact 

Does not finish 
- [ ] feedback form
- [ ] Recollect Recycling Guidelines
- [ ] Related Services
- [ ] 311 Section
- [ ] Update contact svgs
- [ ] Update Map application block with contact data

@benweatherman -- some joplin content needs:
- format all Steps the same —> all in p tags
- send contact data with application block map (see Celine's mock)
- add `<h1>What Else Do I Need to Know?</h1>` to all service page extra content html blocks
- only one contact per service page
- hearing impaired phone number in addition to standard phone number in contact data
- country without underscores in contact data
- contact hours set for all days of week


src/css/base/_variables.scss
```diff
@@ -88,8 +83,5 @@ $coa-color-gray-light: #CBCBCB;
 $coa-color-gray: #919191;
 $coa-color-gray-dark: #363636;
 
-
-
-
-
-
+//magic numbers
```
mateoclarke: 🎩 🐰 ✨ 
https://www.youtube.com/watch?v=aU4pyiB-kq0

src/css/modules/_ApplicationBlock.scss
```diff
@@ -0,0 +1,98 @@
+.coa-ApplicationBlock__title {
+  text-transform: uppercase;
+  text-align: center;
+}
+
+#rCw.coa-ApplicationBlock__Recollect {
```
mateoclarke: I think it would be cool to isolate this Recollect override code in its own file with maybe an explanatory comment at the top.

src/js/modules/ContactAddress.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+//TODO: import correct svg
```
mateoclarke: https://github.com/encharm/Font-Awesome-SVG-PNG/tree/master/black/svg

src/js/modules/ContactHours.js
```diff
@@ -33,50 +36,42 @@ const i18nMessagesWeekdayMap = {
...
-          <table className="usa-table-borderless">
-            <thead className="usa-sr-only">
-              <tr>
-                <th scope="col">Day</th>
-                <th scope="col">Open - Close Hours</th>
+    const hours = this.sort(this.props.hours);
+    return (
+      <div className={`${this.props.className} coa-ContactHours`}>
+        <AirplaneSVG size="20"/>
+        <table className="usa-table-borderless">
```
mateoclarke: Should we copy/🍝  these two classes from USWDS? You're doing a lot of work moving us off USWDS so just wondering why you didn't here too.

`usa-table-borderless`: https://github.com/uswds/uswds/blob/9ff31e9e2c8906246b79d233116ef842cbc5a52c/src/stylesheets/elements/_table.scss#L29
`usa-sr-only`: https://github.com/uswds/uswds/blob/1b16c2dc97c3a38c91f9af1eb5c2694c6d77e773/src/stylesheets/core/_utilities.scss#L52-L69

mateoclarke:
One more comment... I know you said you aren't adding components to Storybook yet because of the usability testing time crunch. But it is still a part of the commitments we've made for this sprint. Can you make an issue with a list of the components that you're refactoring that need to be added to Storybook once we have some breathing room?

src/css/modules/_ApplicationBlock.scss
ifsimicoded: 👍

src/js/modules/ContactAddress.js
ifsimicoded: Thank you

src/js/modules/ContactHours.js
ifsimicoded: I agree we should, I mostly didn't want to deal with it the later it got in the day. We're probably leaning on a lot of their base styles too -- I'll log this in an issue.

ifsimicoded:
@mateoclarke I'm not going to make a new issue for storybooks since it's already part of the existing issue. I just won't close the existing issue until all the components have been added. is ok?

PR #177, opened 2018-03-20
Language Select design v2

mateoclarke:
_Part of https://github.com/cityofaustin/techstack/issues/180_

screen shot 2018-03-20 at 6 28 44 pm

screen shot 2018-03-20 at 6 28 55 pm

### DIFF
--- a/src/js/page_sections/LanguageSelectBar.js
+++ b/src/js/page_sections/LanguageSelectBar.js
@@ -5,86 +5,35 @@ import { SUPPORTED_LANGUAGES } from 'js/i18n/constants';
 import CaretDownSVG from 'js/svg/CaretDown';
 import CaretUpSVG from 'js/svg/CaretUp';
 
-class LanguageSelectBar extends Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      isOpen: false,
-    };
-  }
 
-  handleExpandMenu = () => {
-    this.setState({
-      isOpen: !this.state.isOpen
-    });
-  }
+const selectedLanguage = (lang) => find(SUPPORTED_LANGUAGES, {'code': lang});
 
-  handleCancel = () => {
-    this.setState({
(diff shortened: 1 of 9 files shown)


ifsimicoded:
Is the intent for the links to change position depending on which one is selected. 
screen shot 2018-03-20 at 7 07 46 pm
screen shot 2018-03-20 at 7 07 39 pm

src/js/page_sections/LanguageSelectBar.js
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from 'react';
+import { NavLink } from 'react-static';
+import { find } from 'lodash';
+import { SUPPORTED_LANGUAGES } from 'js/i18n/constants';
+
+const selectedLanguage = (lang) => find(SUPPORTED_LANGUAGES, {'code': lang});
+
+const LanguageSelectBar = ({path, lang}) => (
+  <div className="coa-LanguageSelectBar">
+    <div className="wrapper">
```
ifsimicoded: just an FYI, container-fluid is a class that auto adds 2rems of padding to the left and right of a container div.
I've been using it with wrapper to always ensure there's the same size padding on the left and right of container divs and same size max-widths.
So like
`<div class="container-fluid wrapper"> ...</div>`

mateoclarke: 👍 👍 👍

cthibaultatx:
@mateoclarke seeing your question above and I believe we covered it in issue [#207](https://github.com/cityofaustin/techstack/issues/207). The words should always stay in the same position- they are prioritized by number of people speaking each language at home,

1) Spanish
2) Vietnamese - next in line behind Chinese which we are building later
3) Arabic
4) In last position because it's highest spoken and default.

Important to keep user controls in the same place to reduce a person's memory load

PR #180, opened 2018-03-21
Tile Groups

ifsimicoded:
Tiles
Tile Groups
Section Header
Step fixes
Remove reference to old SectionTitle and RelatedLinks


src/css/modules/_Steps.scss
```diff
@@ -11,7 +11,7 @@
 }
 
 .coa-Steps__list {
-  & > ol {
+  & > ul {
```
benweatherman: Oh snap, this might've been my bad. We should probably keep both `ul` and `ol`, but probably Joplin should be returning `ol` instead of `ul`

src/css/modules/_SectionHeader.scss
```diff
@@ -0,0 +1,12 @@
+.coa_SectionHeader {
+  text-transform: uppercase;
+  text-align: center;
+  &:after {
+    margin: $coa-spacing-medium auto $coa-spacing-xlarge;
+    width: 40px;
```
mateoclarke: When is it chill to have magic numbers like this? For me, I don't mind too much. Would it be better at all to call them out and assign them as sass vars and throw them at the top of the page? idk. seeing it a lot for widths and heights

src/css/modules/_TileGroup.scss
```diff
@@ -0,0 +1,18 @@
...
+  flex-direction: row;
+  flex-wrap: wrap;
+
+  .coa-Tile {
+    flex-grow: 1 ;
+    flex-shrink: 0;
+    flex-basis: 98%;
+    margin: 1%;
+    @include media($coa-small-screen) {
+      flex-basis: 50% - 4%;
```
mateoclarke: is this compensating for margin? perhaps 1% become `$tile-margin`. 
Then line 12 becomes: `flex-basis: 50% - (4 * $tile-margin);`.
Is that any more clear?

src/js/pages/Department.js
```diff
@@ -29,11 +28,11 @@ const Department = ({ department }) => {
           <div className="coa-main__left col-xs-12 col-lg-8">
 
             <div className="coa-section">
-              <SectionTitle title={title}/>
+              <SectionHeader title={title}/>
             </div>
 
             <div className="coa-section">
-              <SectionTitle title="Our Mission"/>
+              <SectionHeader title="Our Mission"/>
```
mateoclarke: should we go ahead and put this text in some i18nMessage object?

src/js/pages/Departments.js
```diff
@@ -13,10 +12,6 @@ const Departments = ({ allDepartments }) => {
   return (
     <div className="wrapper">
       <Hero callout={"All Departments"} />
```
mateoclarke: more i18n object low hanging 🍇

src/js/pages/Service.js
```diff
@@ -44,10 +45,10 @@ const Service = ({ service }) => {
 
         { steps && <Steps steps={steps} /> }
 
-        { (dynamicContent && dynamicContent.length) && (
-          dynamicContent.map((content) => (
+        { !!dynamicContent && (
```
mateoclarke: ‼️

src/css/modules/_Steps.scss
ifsimicoded: truth. both cases would've made sense. I'd leave Joplin for now.

src/js/pages/Department.js
ifsimicoded: prolly, I haven't cared much about dept pages -- so I figured I'd do that work later as that page gets more stable.

src/css/modules/_TileGroup.scss
ifsimicoded: YES will add to next PR

src/css/modules/_SectionHeader.scss
ifsimicoded: mmmm, honestly, not too sure.

PR #184, opened 2018-03-21
Handle empty department contact

benweatherman:
The build was failing due to something like

```
TypeError: Cannot read property 'name' of undefined

    at Contact (/app/dist/static.288553d5.js:1578:22)
```

We need to account for empty contacts


benweatherman:
The deployed version is working, so gonna merge

PR #189, opened 2018-03-22
Parse phone number as JSON

benweatherman:
I also switched it from TDD to TTY because
1. It seems like TTY is preferred in the deaf community http://www.nchearingloss.org/tty.htm
2. TDD is Test Driven Development :smiling_imp:

Part of cityofaustin/techstack#207

### DIFF
--- a/src/js/helpers/cleanData.js
+++ b/src/js/helpers/cleanData.js
@@ -26,6 +26,9 @@ export const cleanContacts = (contacts) => {
 
   return contacts.edges.map(({node: contact}) => {
     let {contact: cleaned} = contact;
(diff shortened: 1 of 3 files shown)


src/js/helpers/cleanData.js
```diff
@@ -26,6 +26,9 @@ export const cleanContacts = (contacts) => {
 
   return contacts.edges.map(({node: contact}) => {
     let {contact: cleaned} = contact;
+    if(cleaned.phone) {
```
benweatherman: oh yes, this is much nicer 🙇

PR #190, opened 2018-03-23
Right to left (rtl) to language wrapper

mateoclarke:
After doing a bit of research about how this is supposed to work, here is few small change that seem to get the desired effect. Here's some things I learned and some caveats for future refactoring...

- HTML elements can have a `dir` attribute and there is a corresponding CSS rule `direction`. Both use the values `rtl` or `ltr`. In this PR, we use a bit of both. 
  - https://stackoverflow.com/questions/34899513/right-to-left-rtl-support-in-react
  - https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir
  - https://developer.mozilla.org/en-US/docs/Web/CSS/direction

- Additionally, there are some React libraries and CSS tooling for more specialized support. But that doesn't seem necessary at this point, unless usability testing shows otherwise.
  - https://github.com/airbnb/react-with-direction
  - https://cssjanus.github.io/

TODO:

- [x] Breadcrumbs
- [x] Menu dropdown fall offscreen
- [x] translate Menu section about site being a work in progress

mar-22-2018 19-23-53

mar-22-2018 19-24-38

_Closes https://github.com/cityofaustin/techstack/issues/145_

### DIFF
--- a/src/css/modules/_ContactDetails.scss
+++ b/src/css/modules/_ContactDetails.scss
@@ -50,3 +50,12 @@
     border: none;
   }
 }
+
+.rtl {
+  .coa-ContactDetails__item {
+    & > :first-child {
+      margin-right: 0;
+      margin-left: $coa-spacing-large;
+    }
+  }  
+}
--- a/src/css/modules/_ContactDetails.scss
+++ b/src/css/modules/_ContactDetails.scss
@@ -51,11 +51,11 @@
   }
 }
 
-.rtl {
+.coa-rtl {
   .coa-ContactDetails__item {
     & > :first-child {
       margin-right: 0;
       margin-left: $coa-spacing-large;
     }
-  }  
+  }
 }
--- a/src/css/modules/_PageBreadcrumbs.scss
+++ b/src/css/modules/_PageBreadcrumbs.scss
@@ -59,7 +59,7 @@
     direction: rtl;
     display: flex;
     flex-direction: row-reverse;
-    justify-content: flex-end;
+    justify-content: center;
     a {
       &:before {
         content: ">";
--- a/src/css/modules/_PageBreadcrumbs.scss
+++ b/src/css/modules/_PageBreadcrumbs.scss
@@ -53,3 +53,31 @@
     }
   }
 }
+
+.rtl {
+  .coa-PageBreadcrumbs {
+    direction: rtl;
+    display: flex;
+    flex-direction: row-reverse;
+    justify-content: flex-end;
+    a {
+      &:before {
+        content: ">";
+        padding-left: $coa-spacing-small;
+        padding-right: 0;
+      }
+    }
+  }
+  .coa-PageBreadcrumbs__title:before {
+    content: "";
+  }
+  @include media($coa-large-screen) {
+    &.coa-PageBreadcrumbs__home {
+      display: inline;
+      &:before {
+        content: "";
+        padding-right: 0;
+      }
+    }
+  }
+}
--- a/src/css/modules/_PageBreadcrumbs.scss
+++ b/src/css/modules/_PageBreadcrumbs.scss
@@ -1,8 +1,10 @@
 .coa-PageBreadcrumbs {
   border-bottom: 1px dashed $coa-color-gray;
+  display: flex;
   font-family: $coa-font-sans;
   font-size: $coa-font-size-xsmall;
   font-weight: $coa-font-semibold;
+  justify-content: center;
   line-height: $coa-line-height-xlarge;
   overflow: hidden;
   text-transform: uppercase;
@@ -54,12 +56,10 @@
   }
 }
 
-.rtl {
+.coa-rtl {
   .coa-PageBreadcrumbs {
     direction: rtl;
-    display: flex;
     flex-direction: row-reverse;
-    justify-content: center;
     a {
       &:before {
         content: ">";
--- a/src/js/components/LanguageWrapper.js
+++ b/src/js/components/LanguageWrapper.js
@@ -111,11 +111,11 @@ class LanguageWrapper extends Component {
     const { lang } = this.state;
     const messages = localeMessages[lang];
     // Right to Left Direction for arabic text
-    const direction = lang === 'ar' ? 'rtl' : 'auto';
+    const direction = lang === 'ar' ? 'rtl' : 'ltr';
 
     return (
       <IntlProvider locale={lang} messages={messages} defaultLocale={DEFAULT_LANG} key={lang}>
-        <div style={{ position: 'relative' }} dir={direction}>
+        <div style={{ position: 'relative' }} dir={direction} className={direction}>
           <a href="#main" className="usa-skipnav">Skip to main content</a>
           <LanguageSelectBar lang={lang} path={this.props.match.params.path || ''}/>
           <Header />
--- a/src/js/components/LanguageWrapper.js
+++ b/src/js/components/LanguageWrapper.js
@@ -3,6 +3,8 @@ import Routes from 'react-static-routes'
 import locale from 'browser-locale'
 import Cookies from 'js-cookie'
 import PropTypes from 'prop-types'
+import { find } from 'lodash';
+
 
 // react-intl i18n
 import { IntlProvider, addLocaleData } from 'react-intl'
@@ -14,7 +16,7 @@ import messages_en from 'js/i18n/locales/en.json';
 import messages_es from 'js/i18n/locales/es.json';
 import messages_vi from 'js/i18n/locales/vi.json';
 import messages_ar from 'js/i18n/locales/ar.json';
-import { SUPPORTED_LANG_CODES, LANG_COOKIE_NAME, LANG_COOKIE_EXPIRES, DEFAULT_LANG } from 'js/i18n/constants'
+import { SUPPORTED_LANGUAGES, SUPPORTED_LANG_CODES, LANG_COOKIE_NAME, LANG_COOKIE_EXPIRES, DEFAULT_LANG } from 'js/i18n/constants'
 addLocaleData([...en, ...es, ...vi, ...ar]);
 const localeMessages = {
   'en': messages_en,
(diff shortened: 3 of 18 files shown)


benweatherman:
beautiful

src/js/components/LanguageWrapper.js
```diff
@@ -110,10 +110,12 @@ class LanguageWrapper extends Component {
   render() {
     const { lang } = this.state;
     const messages = localeMessages[lang];
+    // Right to Left Direction for arabic text
+    const direction = lang === 'ar' ? 'rtl' : 'ltr';
 
     return (
       <IntlProvider locale={lang} messages={messages} defaultLocale={DEFAULT_LANG} key={lang}>
-        <div style={{ position: 'relative' }}>
+        <div style={{ position: 'relative' }} dir={direction} className={direction}>
```
mateoclarke: This is the important part. If our lang is Arabic, throw on the `rtl` class name and `dir` value for everything in the inheritance below. `dir` takes care of a lot of things automatically in the browser. For everything else, we namespace some CSS rules under `.rtl { ... }`

src/js/page_sections/Menu/Submenu.js
```diff
@@ -1,16 +1,23 @@
...
 import ArrowRightSVG from 'js/svg/ArrowRight';
 
-const Submenu = ({id, openSection, theme, handleMenuToggle}) => (
-  <ul className={`coa-Submenu
-      ${ id > 4 ? `coa-Submenu--align-right` : '' }
-      ${ openSection === id ? 'coa-Submenu--open' : '' }
-    `}
+const getSubmenuClassnames = (intl, id, openSection) => {
+  const base = 'coa-Submenu';
+  const shouldAlignRight = (intl.locale === 'ar' && id < 5) ||  (intl.locale !== 'ar' && id > 4);
```
benweatherman: This magic number took me a sec to figure out. Maybe a const could help clear that up (though it's similar to the code that was there, so 👍 for maintaining existing style 😎 )

src/css/modules/_ContactDetails.scss
```diff
@@ -50,3 +50,12 @@
     border: none;
   }
 }
+
+.rtl {
```
ifsimicoded: SUPER MINOR but can we stick to coa prefixes -- so .coa-rtl

src/js/components/LanguageWrapper.js
```diff
@@ -110,10 +110,12 @@ class LanguageWrapper extends Component {
   render() {
     const { lang } = this.state;
     const messages = localeMessages[lang];
+    // Right to Left Direction for arabic text
+    const direction = lang === 'ar' ? 'rtl' : 'ltr';
```
ifsimicoded: can we make direction a property in js/i18n/constants.js file so we don't need inline conditionals. We can pull direction based on the preset constants.

src/js/i18n/locales/ar.json
```diff
@@ -25,7 +25,7 @@
   "Menu.ThreeOneOneMobileListItem.callText": "Call 311",
   "Menu.ThreeOneOneMobileListItem.onlineText": "Submit an Online Request",
   "Menu.PrivacyPolicyListItem.text": "Read About Privacy",
-  "Menu.MobileFooter.text": "Alpha.austin.gov is a work in progress. For the full City of Austin website, visit",
+  "Menu.MobileFooter.text": "Alpha.austin.gov هو موقع إلكتروني جديد والعمل قيد التنفيذ. للحصول على كامل موقع مدينة أوستن، قم بزيارة {citySiteLink}. تعرف على المزيد حول الموقع الإلكتروني الجديد عبر {citySiteLink}.",
```
ifsimicoded: should the last link be {projectSiteLink}? This text looks a lot like line 34 "Home.Secondarycontent.bodytext" but the last link is different.

src/css/modules/_PageBreadcrumbs.scss
```diff
@@ -53,3 +53,31 @@
     }
   }
 }
+
+.rtl {
+  .coa-PageBreadcrumbs {
+    direction: rtl;
+    display: flex;
```
ifsimicoded: all the base styles that could work for left to right or right to left -- can we move them up to
`.coa-PageBreadcrumbs` so the component style behavior works the same for both rtl and ltr -- and it's obvious what is changed specifically for rtl.

so, can you move these two up to the generic .coa-PageBreadcrumbs
`display: flex;
justify-content: center;
`

ifsimicoded: Also -- this might be dumb -- but why both row-reverse and direction rtl?

src/js/i18n/locales/ar.json
```diff
@@ -34,5 +34,6 @@
   "Home.Secondarycontent.bodytext": "Alpha.austin.gov هو موقع إلكتروني جديد والعمل قيد التنفيذ. للحصول على كامل موقع مدينة أوستن، قم بزيارة {citySiteLink}. تعرف على المزيد حول الموقع الإلكتروني الجديد عبر {projectsSiteLink}.",
   "Service.RelatedLinks.SectionHeader": "الاطلاع على الخدمات ذات الصلة",
   "Service.RelatedLinks.Tag": "الخدمة",
-  "Home.RelatedLinks.Tag": "الخدمة"
+  "Home.RelatedLinks.Tag": "الخدمة",
+  "Submenu.workInProgress": "Alpha.austin.gov هو موقع إلكتروني جديد والعمل قيد التنفيذ. للحصول على كامل موقع مدينة أوستن، قم بزيارة {citySiteLink}. تعرف على المزيد حول الموقع الإلكتروني الجديد عبر {citySiteLink}."
```
ifsimicoded: Same here -- should the last link be {projectSiteLink}? This text looks a lot like line 34 "Home.Secondarycontent.bodytext" but the last link is different

src/js/page_sections/Menu/Menu.js
```diff
@@ -174,7 +171,13 @@ const PrivacyPolicyListItem = ({intl}) => (
 const MobileFooter = ({intl}) => (
   <div className="coa-Menu__mobile-footer">
     <p className="coa-Menu__footer-text d-lg-none">
-      {intl.formatMessage(i18nMessages.footer)}<ExternalLink to="https://austintexas.gov">austintexas.gov</ExternalLink>.
+      <FormattedMessage
+        id="Menu.MobileFooter.text"
+        defaultMessage="Alpha.austin.gov is a work in progress. For the full City of Austin website, visit {citySiteLink}."
```
ifsimicoded: so a few things.
1. by changing the format here by making the link a parameter and part of the formatted message, you're ALSO changing the display of the links in all languages. you'll have to update the other language json files like you did by updating the en.json to include ` {citySiteLink}` as the end of the string.
2. this mobile menu footer text looks different than the arabic version which shows two links? The format should be the same.
3. Awhen you use FormattedMessage, it throws everything in a container span.

src/js/i18n/locales/en.json
```diff
@@ -25,7 +25,7 @@
   "Menu.ThreeOneOneMobileListItem.callText": "Call 311",
   "Menu.ThreeOneOneMobileListItem.onlineText": "Submit an Online Request",
   "Menu.PrivacyPolicyListItem.text": "Read About Privacy",
-  "Menu.MobileFooter.text": "Alpha.austin.gov is a work in progress. For the full City of Austin website, visit",
+  "Menu.MobileFooter.text": "Alpha.austin.gov is a work in progress. For the full City of Austin website, visit {citySiteLink}.",
```
ifsimicoded: Ok, so it looks like Menu.MobileFooter.text should only include citySiteLink and not both city and project. Is showing different content for arabic vs the other languages just a stop gap for arabic testing tomorrow?

src/js/page_sections/Menu/Submenu.js
ifsimicoded: I second the constant part of ^

src/js/i18n/locales/en.json
```diff
@@ -34,5 +34,6 @@
   "Home.Secondarycontent.bodytext": "Alpha.austin.gov is a new website and a work in progress. For the full City of Austin website, visit  {citySiteLink}. Learn more about the new website at {projectsSiteLink}.",
   "Service.RelatedLinks.SectionHeader": "Check out related services",
   "Service.RelatedLinks.Tag": "Service",
-  "Home.RelatedLinks.Tag": "Service"
+  "Home.RelatedLinks.Tag": "Service",
+  "Submenu.workInProgress": "Alpha.austin.gov is a work in progress. For the full City of Austin website, visit {citySiteLink}."
```
ifsimicoded: We'll have to add Submenu.workInProgress for all language json files. This we can grab this from the first half of Home.Secondarycontent.bodytext? It might get weird in vi, I dunno.

src/js/page_sections/Menu/Menu.js
```diff
@@ -108,6 +104,7 @@ class Menu extends Component {
               theme={theme}
               handleMenuToggle={this.props.toggleMenu}
               handleSublistToggle={this.toggleSublist}
+              intl={intl}
```
ifsimicoded: instead of passing intl around as a prop can we just inject it where we need it like: 
`const superI18ncomponent = injectIntl(({someprop, intl}) => ( ... ) `

ifsimicoded: In all fairness, I probably set pretty bad precedent for this in Menu.js. Instead of `<AirportMobileListItem intl={intl} />` I should have ` const AirportMobileListItem = injectIntl((props) => ( ... ))`

src/js/page_sections/Menu/MenuItem.js
```diff
@@ -40,6 +40,7 @@ const MenuItem = ({theme, id, openSection, handleSublistToggle, handleMenuToggle
           openSection={openSection}
           theme={theme}
           handleMenuToggle={handleMenuToggle}
+          intl={intl}
```
ifsimicoded: same here, can we inject intl where needed via wrapping the component with injectIntl like in the above comment.

src/js/page_sections/Menu/Submenu.js
```diff
@@ -1,16 +1,23 @@
...
+  const alignRight = shouldAlignRight ? 'coa-Submenu--align-right' : '';
+  const open = openSection === id ? 'coa-Submenu--open' : '';
+
+  return `${base} ${alignRight} ${open}`;
+}
+
+const Submenu = ({id, openSection, theme, handleMenuToggle, intl}) => (
+  <ul className={getSubmenuClassnames(intl, id, openSection)}
```
ifsimicoded: same let's inject intl, instead of passing it as an arg.

src/css/modules/_ContactDetails.scss
mateoclarke: suuuuuure

src/js/components/LanguageWrapper.js
mateoclarke: i think i know what you mean here. will do.

src/js/i18n/locales/ar.json
mateoclarke: It's actually supposed to be a shortened version of "Home.Secondarycontent.bodytext". Good catch.

src/css/modules/_PageBreadcrumbs.scss
mateoclarke: So I think I got confused here on the styling because the breadcrumbs aren't pulling in the arabic translations yet which makes them act weird. Once those are coming in, I think the ordering will behave better, but we might need some refactoring. I'll log an issue.

src/js/i18n/locales/ar.json
mateoclarke: same as above

src/js/page_sections/Menu/Menu.js
mateoclarke: Since the two `<FormattedMessage>`s are the same, any issue if they just use the same id?

src/css/modules/_PageBreadcrumbs.scss
ifsimicoded: 👍

PR #195, opened 2018-03-24
Topic Breadcrumb updates

ifsimicoded:
- use joplin delivered topic data in breadcrumbs to get translated content
- use ids instead of slugs for navigation until slugs are implemented
- add slugs to joplin queries to prep for future slug usage

### DIFF
--- a/src/js/modules/PageBreadcrumbs.js
+++ b/src/js/modules/PageBreadcrumbs.js
@@ -5,13 +5,13 @@ const PageBreadcrumbs = ({ title, order, ...rest }) => {
 
   const breadcrumbs = order.map(breadcrumb => ({
     className: breadcrumb,
-    title: rest[breadcrumb].title,
-    slug: rest[breadcrumb].slug,
+    text: rest[breadcrumb].text,
+    slug: `/${breadcrumb}s/${rest[breadcrumb].id}`,
   }));
 
   breadcrumbs.unshift({
     className: 'home',
-    title: 'Home',
+    text: 'Home',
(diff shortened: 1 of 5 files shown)


src/js/modules/PageBreadcrumbs.js
```diff
@@ -5,13 +5,13 @@ const PageBreadcrumbs = ({ title, order, ...rest }) => {
 
   const breadcrumbs = order.map(breadcrumb => ({
     className: breadcrumb,
-    title: rest[breadcrumb].title,
-    slug: rest[breadcrumb].slug,
+    text: rest[breadcrumb].text,
+    slug: `/${breadcrumb}s/${rest[breadcrumb].id}`,
```
benweatherman: Oh wait, can we use slug here? Is this for theme or topic?

benweatherman: Oh wait, can we use slug here? Is this for theme or topic?

ifsimicoded: it's the same for both topic and theme since I'm just loo[ing through all links in the breadcrumb list.

benweatherman: Why move to IDs? I thought we were moving towards slugs for all.

ifsimicoded: Not moving to Id's -- I'm preserving the ids we had before so navigation still works. Our queries currently fetch topic page data via id. I think that's how the Joplin queries are set up for topic data.

PR #196, opened 2018-03-24
Recollect Tabbing and HashLink

ifsimicoded:
Added option for Tabbed Recollect -- Joplin data should be formatted as below 

```
"dynamicContent": [
        {
          "id": "a4e532e1-8113-4552-b2ba-4df23c1edbf8",
          "type": "recollect_block",
          "value": null
        },
        ...
]
```
Jump Links to Recollect should look like 
`<a href="#HashLink-Recollect">Go To Recollect</a>`

@cthibaultatx I'm sorry this look horrible. We can style this as you see fit.
screen shot 2018-03-23 at 9 51 20 pm

### DIFF
--- a/src/js/modules/Recollect.js
+++ b/src/js/modules/Recollect.js
@@ -6,11 +6,11 @@ class Recollect extends Component {
   handleScriptLoad() {
     // Recollect is a third party script that isn't an import-able node module
     // rCw is the required container id to ensure styles are shown correctly
-    let loader = new window.Recollect.Widget.Loader({
-        area: "Austin",
-        name: this.props.name || 'calendar',
-        container: "#rCw"
-      });
+    const options = Object.assign({
+      area: "Austin",
+      container: "#rCw"
+    }, this.props.options);
+    let loader = new window.Recollect.Widget.Loader(options);
     loader.load();;
   }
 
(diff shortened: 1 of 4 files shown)


src/js/modules/Recollect.js
```diff
@@ -6,11 +6,11 @@ class Recollect extends Component {
   handleScriptLoad() {
     // Recollect is a third party script that isn't an import-able node module
     // rCw is the required container id to ensure styles are shown correctly
-    let loader = new window.Recollect.Widget.Loader({
-        area: "Austin",
-        name: this.props.name || 'calendar',
-        container: "#rCw"
-      });
+    const options = Object.assign({
```
mateoclarke: I'm interested to know what motivated this refactoring. Guessing it helps with multiple instances on a page. maybe?

ifsimicoded:
@cthibaultatx couldn't take how awful it was so I took a stab at making it less hideous.
screen shot 2018-03-24 at 1 19 03 am

src/js/modules/Recollect.js
ifsimicoded: Yea, not multiple instances per page, but it was the simplest refactor, I thought, to enable adding a tabbed instance.

cthibaultatx:
Oh geez! The buttons @ifsimicoded I'd forgotten to write that down and work on it. I would make the box around the active tab similar to the secondary button styling, I think. I'll start a new issue for that but it can go in the backlog for now.

Issue #197, opened 2018-03-24
Refactor Breadcrumbs to pull in data from query and refactor styling for Arabic


mateoclarke:
https://github.com/cityofaustin/janis/pull/190#discussion_r176885905

ifsimicoded:
The breadcrumbs pull from Joplin except theme pages as I don't think those have been built in Joplin.

PR #200, opened 2018-03-24
breadcrumb translate home

ifsimicoded:
- addresses more of #197 by translating the word 'Home' 
- keeps the language select bar on the right side regardless of rtl or ltr content so transitioning from language to language is easier

### DIFF
--- a/src/css/page_sections/_LanguageSelectBar.scss
+++ b/src/css/page_sections/_LanguageSelectBar.scss
@@ -18,6 +18,12 @@
   }
 }
 
+.coa-rtl {
+  .coa-LanguageSelectBar__language-list {
+    flex-direction: row-reverse;
+  }
+}
+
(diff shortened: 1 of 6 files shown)


src/css/page_sections/_LanguageSelectBar.scss
```diff
@@ -18,6 +18,12 @@
   }
 }
 
+.coa-rtl {
+  .coa-LanguageSelectBar__language-list {
+    flex-direction: row-reverse;
```
mateoclarke: I was going to wait and see how this worked out in usability testing first before changing. I asked @cthibaultatx on Friday about this and she mentioned that there might be imbalance visually if the language select stayed on the same side. Either way, maybe its worth asking again now that its implemented this way? 🤷‍♂️

PR #203, opened 2018-03-27
Sd somecleanuponmenusandbreadcrumbs

ifsimicoded:
- minor Breadcrumb cleanup
- fix nav menu bug where first click doesn't register (make click target include padding)
- other nav menu clean up (inject intl where used, don't pass as prop)

### DIFF
--- a/src/css/page_sections/Menu/_Menu.scss
+++ b/src/css/page_sections/Menu/_Menu.scss
@@ -64,7 +64,7 @@
   font-size: $coa-font-size-xsmall;
 }
 
-.coa-Menu__click-overlay {
+.coa-Menu__close-submenu-click-target {
   display: none;
   position: fixed;
   width: 100%;
--- a/src/css/page_sections/Menu/_Menu.scss
+++ b/src/css/page_sections/Menu/_Menu.scss
@@ -21,18 +21,24 @@
   }
 }
 
+.coa-Menu--open {
+  left: 0;
+  @include media($coa-large-screen) {
+    left: auto;
+  }
+}
+
 .coa-Menu__list {
   list-style: none;
   @include media($coa-large-screen) {
     display: flex;
     justify-content: space-between;
   }
-}
-
-.coa-Menu--open {
-  left: 0;
-  @include media($coa-large-screen) {
-    left: auto;
+  & > li {
+    margin: 0;
+    @include media($coa-large-screen) {
+      position: relative;
+    }
   }
 }
 
--- a/src/css/page_sections/Menu/_Menu.scss
+++ b/src/css/page_sections/Menu/_Menu.scss
@@ -9,6 +9,7 @@
   width: 100vw;
   z-index: 2;
   @include media($coa-large-screen) {
+    background: none;
     bottom: auto;
     left: auto;
     margin: $coa-space-level5 auto 0;
@@ -16,6 +17,7 @@
     position: static;
     top: auto;
     width: 100%;
+    z-index: auto;
   }
 }
 
@@ -54,14 +56,4 @@
 .coa-Menu__footer-text {
   padding: $coa-space-level3 0;
   font-size: $coa-font-size-xsmall;
-}
-
-.coa-Menu__overlay {
-  position: fixed;
-  width: 100%;
-  height: 100%;
-  z-index: 1;
-  top: 0;
-  left: 0;
-  opacity: 0;
 }
--- a/src/css/page_sections/Menu/_Menu.scss
+++ b/src/css/page_sections/Menu/_Menu.scss
@@ -14,10 +14,10 @@
     left: auto;
     margin: $coa-space-level5 auto 0;
     overflow-y: visible;
-    position: static;
+    position: relative;
     top: auto;
     width: 100%;
-    z-index: auto;
+    z-index: 2;
   }
 }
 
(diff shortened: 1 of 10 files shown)


src/js/page_sections/Menu/Submenu.js
```diff
@@ -1,66 +1,80 @@
...
-  const shouldAlignRight = arabicRightMenuItems ||  nonArabicRightMenuItems;
-  const alignRight = shouldAlignRight ? 'coa-Submenu--align-right' : '';
-  const open = openSection === id ? 'coa-Submenu--open' : '';
+const SUBMENU_THRESHOLD_ALIGNRIGHT_RTL = 5;
+const SUBMENU_THRESHOLD_ALIGNRIGHT_LTR = 4;
 
-  return `${base} ${alignRight} ${open}`;
+const isAlignedRight = (direction, id) => {
+  return (direction === 'rtl') ? id < SUBMENU_THRESHOLD_ALIGNRIGHT_RTL : id > SUBMENU_THRESHOLD_ALIGNRIGHT_LTR;
+};
```
mateoclarke: this is better than what I had been doing. 👏

src/js/page_sections/Menu/Menu.js
```diff
@@ -123,51 +138,57 @@ class Menu extends Component {
...
+    </div>
   </li>
-)
-
-const ThreeOneOneMobileListItem = ({intl}) => (
-  <li className="coa-MenuItem coa-MenuItem--flex coa-MenuItem--small d-lg-none">
-    <a href="tel:311">
-      {intl.formatMessage(i18nMessages.call311)}
-    </a>
-    &nbsp;or&nbsp;
```
mateoclarke: we need these. or some padding. JSX doesn't keep spaces between elements

src/css/page_sections/Menu/_Menu.scss
```diff
@@ -9,13 +9,22 @@
...
+    background: none;
     bottom: auto;
     left: auto;
     margin: $coa-space-level5 auto 0;
     overflow-y: visible;
-    position: static;
+    position: relative;
     top: auto;
     width: 100%;
+    z-index: 2;
```
mateoclarke: maybe we need to get more sophisticated with how we do z-index. Maybe this is just an issue to log for later.
https://css-tricks.com/handling-z-index/

src/css/page_sections/Menu/_MenuItem.scss
```diff
@@ -1,91 +1,88 @@
...
 
 .coa-MenuItem--small {
   font-size: $coa-font-size-xsmall;
-}
+  white-space: nowrap;
 
-.coa-MenuItem--home > a {
-  color: $coa-color-orange;
-  text-decoration: none;
+  &.coa-MenuItem--home > a {
```
mateoclarke: I don't think this should be nested unless is needs to be. It just adds extra specificity and makes it harder to overwrite the rule if needed

src/js/page_sections/Header.js
```diff
@@ -59,7 +59,7 @@ class Header extends Component {
               </div>
             </div>
             <div className="col-xs-6 coa-Header__right-controls">
-              <div className="d-none d-md-block">
+              <div className="d-none d-lg-block">
```
mateoclarke: These links show in the Sketch mocks on medium screens. If we are going to change them, let's make sure that is @cthibaultatx's intent to stray from what was designed.

src/js/page_sections/Menu/Menu.js
```diff
@@ -62,59 +62,74 @@ class Menu extends Component {
...
-            <MobileFooter intl={intl} />
+            <PrivacyPolicyMenuItem handleToggleAllMenus={this.toggleAllMenus} />
+            <MobileFooter />
           </ul>
         </nav>
         {
-          !!this.state.openSection && (
-            <div className="coa-Menu__overlay"
-              onClick={() => this.setState({ openSection: null })}
+          !!this.state.openSubmenuId && (
```
mateoclarke: maybe we add a comment since it might not be clear what this overlay is for?

src/js/page_sections/Menu/Submenu.js
```diff
@@ -1,66 +1,79 @@
...
+
+const Submenu = ({id, theme, isSubmenuOpen, handleToggleAllMenus, intl}) => {
+
+  const langObj = find(SUPPORTED_LANGUAGES, {'code': intl.locale});
+  return (
+    <ul className={`coa-Submenu
+      ${isAlignedRight(langObj.direction, id) ? 'coa-Submenu--align-right' : ''}
+      ${isSubmenuOpen ? 'coa-Submenu--open' : ''}`
```
mateoclarke: As discussed, I prefer keeping classNames free of lots of logic. I normally do this with a separate function like `getSubmenuClassnames()`. You've mentioned that you like the way the 'classnames' npm package works. Maybe we add that as a task for refactoring? https://www.npmjs.com/package/classnames

src/css/page_sections/Menu/_MenuItem.scss
ifsimicoded: 🌮

src/css/page_sections/Menu/_Menu.scss
ifsimicoded: 💅

src/js/page_sections/Header.js
ifsimicoded: 📟

PR #204, opened 2018-03-27
Topic page

mateoclarke:
Dependent on Joplin updates: https://github.com/cityofaustin/joplin/pull/109

mar-26-2018 21-24-52

screen shot 2018-03-26 at 9 22 22 pm

screen shot 2018-03-26 at 9 22 03 pm

### DIFF
--- a/src/js/pages/Topic.js
+++ b/src/js/pages/Topic.js
@@ -2,41 +2,50 @@ import React from 'react';
 import { withRouteData } from 'react-static';
 import { get } from 'lodash';
 import { cleanServiceLinks } from 'js/helpers/cleanData';
+import PropTypes from 'prop-types';
 
 // TODO: this jsonFileData is temporary. Add it to Wagtail API
 import jsonFileData from '__tmpdata/pages';
-import Hero from 'js/modules/Hero';
 import TileGroup from 'js/modules/TileGroup';
+import PageBreadcrumbs from 'js/modules/PageBreadcrumbs';
+import PageHeader from 'js/modules/PageHeader';
+import SectionHeader from 'js/modules/SectionHeader';
 import FormFeedback from 'js/page_sections/FormFeedback';
 import ThreeOneOne from 'js/page_sections/ThreeOneOne';
 
 const Topic = ({ topic }) => {
   const title = get(topic, "text", null);
-  const body = get(topic, "description", null);
+  const callToAction = get(topic, "callToAction", null);
+  const description = get(topic, "description", null);
   const links = get(topic, "services", null);
+  const theme = get(topic, "theme", null);
   const relatedLinks = cleanServiceLinks(links);
   const services311 = get(jsonFileData, "services311", null);
 
   return (
     <div>
-      <Hero callout={title} />
+      <PageBreadcrumbs title={title} order={['topic']} topic={theme} />
 
-      <div className="wrapper">
-        <div className="coa-main__body" dangerouslySetInnerHTML={{__html: body}} />
+      <div className="wrapper wrapper--sm container-fluid">
+        <PageHeader title={title} description={description} />
+        <SectionHeader title={callToAction} />
       </div>
 
       <TileGroup tiles={relatedLinks} />
-
-      <div className="coa-section coa-section--lightgrey">
-        <div className="wrapper">
-          <FormFeedback />
-          <a className="coa-section__link" href="#">Return to Top</a>
-        </div>
-      </div>
 
       <ThreeOneOne services311={services311} />
     </div>
   )
 }
+
+Topic.propTypes = {
+  topic: PropTypes.shape({
+    text: PropTypes.string,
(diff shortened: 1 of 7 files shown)


src/js/pages/Topic.js
```diff
@@ -2,41 +2,50 @@ import React from 'react';
 import { withRouteData } from 'react-static';
 import { get } from 'lodash';
 import { cleanServiceLinks } from 'js/helpers/cleanData';
+import PropTypes from 'prop-types';
```
ifsimicoded: 👏 Should we talk about adding propTypes for next sprint? Seems like a good house-cleaning/code-usability thing like storybooks.

src/js/pages/Topic.js
```diff
@@ -2,41 +2,50 @@ import React from 'react';
...
-import Hero from 'js/modules/Hero';
 import TileGroup from 'js/modules/TileGroup';
+import PageBreadcrumbs from 'js/modules/PageBreadcrumbs';
+import PageHeader from 'js/modules/PageHeader';
+import SectionHeader from 'js/modules/SectionHeader';
 import FormFeedback from 'js/page_sections/FormFeedback';
 import ThreeOneOne from 'js/page_sections/ThreeOneOne';
 
 const Topic = ({ topic }) => {
   const title = get(topic, "text", null);
```
ifsimicoded: Super minor, but I went back to destructuring like you recommended elsewhere, when the object isn't terribly nested so we could remove lodash as a dependency where possible. 
how do you feel about that?
const { title, callToAction, ... } = topic?

src/js/pages/Topic.js
```diff
@@ -2,41 +2,50 @@ import React from 'react';
...
+  const description = get(topic, "description", null);
   const links = get(topic, "services", null);
+  const theme = get(topic, "theme", null);
   const relatedLinks = cleanServiceLinks(links);
   const services311 = get(jsonFileData, "services311", null);
 
   return (
     <div>
-      <Hero callout={title} />
+      <PageBreadcrumbs title={title} order={['topic']} topic={theme} />
```
ifsimicoded: any reason your changing the name from theme to topic?
<PageBreadcrumbs title={title} order={['theme']} theme={theme} />

src/js/pages/Topic.js
```diff
@@ -2,41 +2,50 @@ import React from 'react';
...
-          <a className="coa-section__link" href="#">Return to Top</a>
-        </div>
-      </div>
-
       <ThreeOneOne services311={services311} />
     </div>
   )
 }
 
+Topic.propTypes = {
```
ifsimicoded: 🌈

src/js/pages/Topic.js
```diff
@@ -2,41 +2,50 @@ import React from 'react';
...
+      <PageBreadcrumbs title={title} order={['topic']} topic={theme} />
 
-      <div className="wrapper">
-        <div className="coa-main__body" dangerouslySetInnerHTML={{__html: body}} />
+      <div className="wrapper wrapper--sm container-fluid">
+        <PageHeader title={title} description={description} />
+        <SectionHeader title={callToAction} />
       </div>
 
       <TileGroup tiles={relatedLinks} />
```
ifsimicoded: you can pass in a tag prop so the word 'service' shows up.
```
const i18nMessages = defineMessages({
 ...
  serviceRelatedlinksTag: {
    id: 'Service.RelatedLinks.Tag',
    defaultMessage: 'Service',
  }
});
<TileGroup tiles={cleanedRelated} tag={intl.formatMessage(i18nMessages.serviceRelatedlinksTag)} />
```

ifsimicoded: I think the TileGroup also needs a wrapper so it doesn't go end to end.
```
<div className="wrapper container-fluid">
<TileGroup tiles={relatedLinks} tag={intl.formatMessage(i18nMessages.serviceRelatedlinksTag)} />
</div>
```

src/js/pages/Topic.js
mateoclarke: es6 ftw!!!!

src/js/pages/Topic.js
mateoclarke: Breadcrumbs needs more refactoring as we discussed the ordering thing is clever, but could be improved by passing in an array of object where the keys are the classname and the value is the object in the order props. Also, the css is pretty coupled to the one use case in the Services page. We could try to make the classnames abstracted away form topics, themes, titles and use something like child, parent, grandparent. Or we could be clever with nth and last-child psuedo- selectors. 😓

src/js/pages/Topic.js
mateoclarke: yes 😄

src/js/pages/Topic.js
mateoclarke: yup, let's commit to adding them into components as we work on them. Leave it better than you found it, right?

PR #205, opened 2018-03-27
Add Theme pages

mateoclarke:
_Closes cityofaustin/techstack#98_
_Closes cityofaustin/techstack#184_

## With only 1 topic group
screen shot 2018-03-30 at 6 04 07 pm

## With more than 1 topic group side by side
screen shot 2018-03-30 at 5 58 13 pm

## With more than 1 topic group in smaller viewport stacked
screen shot 2018-03-30 at 5 58 35 pm


src/js/modules/TileGroupSet.js
```diff
@@ -0,0 +1,49 @@
...
+import TileGroup from 'js/modules/TileGroup';
+
+
+const TileGroupSet = ({ groups, tileKey, groupTitleSubPath }) =>  {
+
+  // gets the number of topics that have active services
+  const countOfGroups = groups.edges
+    .map((g) => g.node.services.edges.length > 0)
+    .reduce((total, amount) => amount ? total + 1 : total, 0);
+  const shouldTilesStackVertical = countOfGroups > 1;
```
mateoclarke: This whole section that provides the logic for if we display the tiles stacked or side-by-side felt hacky but I couldn't figure out anything more elegant. As designed, it depends on whether the Theme has more than one Topic with active Service pages.  This gets passed into the TileGroup as the `direction` props which will be either `'column'` or `false`. v open to suggestions on how to make this better.

Dockerfile
```diff
@@ -8,11 +8,11 @@ ENV NODE_PATH=src
 
 COPY yarn.lock /app/yarn.lock
 COPY package.json /app/package.json
+RUN yarn
```
ifsimicoded: 🙏

src/css/modules/_TileGroup.scss
```diff
@@ -16,3 +12,28 @@
     }
   }
 }
+
+.coa-TileGroup--border {
+  border: 4px solid $coa-color-blue;
+  border-radius: 6px;
```
ifsimicoded: super picky -- can we make these more tile group variables. so if we have border-radiuses elsewhere we can keep consistent.

src/js/modules/SectionHeader.js
```diff
@@ -1,8 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import ArrowRight from 'js/svg/ArrowRight';
 
-const SectionHeader = ({ title }) => (
-  <h2 className="coa_SectionHeader">{title}</h2>
+
+const SectionHeader = ({ title, arrow }) => (
```
ifsimicoded: also picky -- for boolean props like arrow, can we make the name obvious that the prop is a bool. Like, hasArrow?

src/js/modules/SectionHeader.js
```diff
@@ -1,8 +1,19 @@
...
+  <div>
+    <h2 className="coa_SectionHeader">
+      {title}
+      {arrow && (
+        <span className="coa_SectionHeader--arrow">
```
ifsimicoded: more picky -- should coa_SectionHeader--arrow be coa_SectionHeader__arrow, since the arrow is an element within the SectionHeader block and not a modifier?

src/js/modules/TileGroup.js
```diff
@@ -1,19 +1,49 @@
...
+
 
-const TileGroup = ({ tiles, tag }) => (
-  <div className="coa-TileGroup">
-  {
-    tiles.map(({ url, text }, index) =>
+const TileGroup = ({ tiles, tag, border, title, description, titlePath, direction }) => (
+  <div className={`
+    col-xs-12 ${direction ? 'col-md-6 col-lg-3' : ''}
+    coa-TileGroup ${border ? 'coa-TileGroup--border' : ''}
```
ifsimicoded: same picky, can we make border, hasBorder?

src/js/pages/Theme.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { withRouteData } from 'react-static';
+import PropTypes from 'prop-types';
+
+// TODO: this jsonFileData is temporary. Add it to Wagtail API
+import jsonFileData from '__tmpdata/pages';
+import TileGroupSet from 'js/modules/TileGroupSet';
+import PageBreadcrumbs from 'js/modules/PageBreadcrumbs';
```
ifsimicoded: kinda weird that we don't use breadcrumbs, but since we don't, you can kill the import.

src/js/pages/Themes.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { withRouteData } from 'react-static';
+
+import Hero from 'js/modules/Hero';
+import TileGroup from 'js/modules/TileGroup';
+
+const Themes = ({ allThemes }) => {
+  const links = allThemes.edges.map(theme => ({
```
ifsimicoded: we have a very similar method in src/js/helpers/cleanData.js --> cleanServiceLinks
maybe we can make that method more generic and share?

src/js/pages/Topics.js
```diff
@@ -6,7 +6,7 @@ import TileGroup from 'js/modules/TileGroup';
 
 const Topics = ({ allTopics }) => {
   const links = allTopics.edges.map(topic => ({
-    url: `/topics/${topic.node.id}`,
+    url: `/topics/${topic.node.slug}`,
```
ifsimicoded: same as above, we have a very similar method that does this mapping in cleanData.js

static.config.js
```diff
@@ -1,191 +1,163 @@
...
 
 // QUERIES
 import allServicePagesQuery from 'js/queries/allServicePagesQuery';
 import allTopicPagesQuery from 'js/queries/allTopicPagesQuery';
 import allDepartmentPagesQuery from 'js/queries/allDepartmentPagesQuery';
 import topServicesQuery from 'js/queries/topServicesQuery';
+import allThemesQuery from 'js/queries/allThemesQuery';
+
+
+const createGraphQLClientsByLang = (lang) => {
```
ifsimicoded: confused -- why are these changes showing in this PR? Wasn't this part of Ben's PR https://github.com/cityofaustin/janis/pull/206

benweatherman: Mine was a PR to his PR

ifsimicoded: OOOoooOOooOOOoOOOOOOoooOOo

src/js/modules/TileGroupSet.js
ifsimicoded: This is gonna take me a minute to wrap my head around. and I don't wanna hold up moving forward and merging. Lemme pull this locally and review and make an issue if I can come up with something else.

src/js/pages/Themes.js
mateoclarke: i tried that at first but I felt like I was passing too many params around then got lazy. Logged it as a tech debt issue so its on the radar to refactor https://github.com/cityofaustin/techstack/issues/244

PR #212, opened 2018-04-03
Breadcrumb makeover

mateoclarke:
_Closes https://github.com/cityofaustin/techstack/issues/221_

### DIFF
--- a/src/js/pages/Service.js
+++ b/src/js/pages/Service.js
@@ -29,11 +29,15 @@ const i18nMessages = defineMessages({
 });
 
 const Service = ({ service, intl }) => {
-  const { image, title, slug, topic, steps, dynamicContent, additionalContent, contacts, related } = service;
+  const {
+    image, title, slug, topic, steps, dynamicContent, additionalContent,
+    contacts, related
+  } = service;
+  
+  const { theme } = topic;
 
   //TODO: data below should be sourced as above
-  const { servicepage, services311 } = jsonFileData;
-  const { theme } = servicepage;
+  const { services311 } = jsonFileData;
 
   //TODO: clean data where sourced
   const contact = cleanContacts(contacts)[0];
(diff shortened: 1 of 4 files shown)


benweatherman:
Doh, totally forgot about this before. Can we also remove this now?

https://github.com/cityofaustin/janis/blob/f818d9b7084d750302ea32d5b9e871db21caa7ca/src/__tmpdata/pages.json#L9-L14

mateoclarke:
@benweatherman, we're still using that file for the mock 311 data. But good call that we can start removing as much as possible.

src/js/pages/Service.js
```diff
@@ -29,11 +29,15 @@ const i18nMessages = defineMessages({
 });
 
 const Service = ({ service, intl }) => {
-  const { image, title, slug, topic, steps, dynamicContent, additionalContent, contacts, related } = service;
+  const {
+    image, title, slug, topic, steps, dynamicContent, additionalContent,
+    contacts, related
+  } = service;
+
+  const { theme } = topic;
```
ifsimicoded: will a topic only have one parent theme? I forget the relationships. Maybe a service can have multiple topics?

src/js/pages/Service.js
```diff
@@ -42,7 +46,11 @@ const Service = ({ service, intl }) => {
   return (
     <div>
       <PageBanner image={image} />
-      <PageBreadcrumbs title={title} order={['theme', 'topic']} theme={theme} topic={topic} />
+      <PageBreadcrumbs
+        grandparent={{...theme, subpath: 'themes'}}
```
ifsimicoded: did we land on having the CMS pass in the full path? If so, did we make a task for that?

src/js/queries/allServicePagesQuery.js
```diff
@@ -10,12 +10,16 @@ const allServicePagesQuery = `
             id
             slug
             text
+            theme {
```
ifsimicoded: same question as above, I forget the topic to theme relationship. 1 to 1? If topics can have multiple themes, this may get goofy. Same if services can be under multiple topics, breadcrumbs can get wild.

ifsimicoded:
too late gif-downsized_large

PR #214, opened 2018-04-04
Moar storybookz

mateoclarke:
_Closes https://github.com/cityofaustin/techstack/issues/222_

Storybook updates to:
- Menu
- MenuItem
- HeroHome

Storybook additions of:
- PageHeader
- SectionHeader
- Tiles
  - TileGroupSet
  - TileGroup
  - Tile
- ThreeOneOne
- PageBreadcrumbs
- PageBanner
- SecondaryContentBanner
- Steps
- StaticMap
- Recollect (doesn't load inside Storybook's iframe, but still can see our custom skinning)
- ListLink
- ~i18nLink~ (logic is dependent on lang codes being in the url)
- ~i18nNavLink~ (logic is dependent on lang codes being in the url)
- HtmlFromAdmin
- Contact
  - ContactDetails
  - Phone
  - Address
  - Email
  - Hours

Other notes on things I did...
  - Deleted unused components: 
    - Hero
    - GlobalSearch
  - Reorganized Contact components into one folder.
  - for now, I decided to skip the ApplicationBlock component since it is just a wrapper case statement that renders either Recollect or StaticMap. If we wanted to add some testing or snapshots around these, it might be good to come back and include. Otherwise, nah.
  - I created two subfolders under stories, `images` and `static_data`. This way, Storybook can run independently of Joplin. This seems like the right thing, but there are also ways you can make async requests from storybooks for a backend API if we needed.


src/stories/Steps.stories.js
```diff
@@ -0,0 +1,27 @@
...
+    <li><p>Drop off your items at the Recycle and ReUse Center at 2514 Business Center Dr, Austin, TX 78744.</p></li>
+  </ul>
+`
+
+storiesOf('Steps', module)
+  .addDecorator(checkA11y)
+  .add('Steps', () => (
+    <IntlProvider locale="en">
+      <div className="wrapper container-fluid">
+        <Steps steps={stepsHtml} />
```
mateoclarke: @ifsimicoded, what would you think about changing the name of this prop to be `stepsHtml`, `stepsUnorderedListHtml` or something like that to signal to users of the component what the Step expects?

src/stories/ListLink.stories.js
```diff
@@ -0,0 +1,30 @@
...
+      <ListLink
+        url={'https://austintexas.gov'}
+        text={'Current City of Austin website'}
+      />
+    </div>
+  ))
+  .add('dark theme', () => (
+    <div className="coa-ThreeOneOne wrapper container-fluid">
+      <div className="coa-ThreeOneOne__listlinks">
+        <div className="coa-ListLink">
```
mateoclarke: Wondering if there is something we should do to improve this. The ListLink component is dependent on css classnaming in its parents in order to change it's styling/theme. In order to illustrate how it looks as currently used in the 311 component, I had to wrap it in some divs with classes. Added a story for the default vs the "dark theme"

ifsimicoded: good call

ifsimicoded:
WOAH. since you've done a lot of work here, should we think about having these hosted? We could use github pages on janis for now it seems.
https://storybook.js.org/basics/exporting-storybook/

read reading rainbow gif by levar burton kids-downsized_large

src/js/modules/Contact/Address.js
```diff
@@ -0,0 +1,27 @@
...
+      <span>{location.street}</span>
+      <span>{location.city}, {location.state} {location.zip}</span>
+    </div>
+  </div>
+);
+
+Address.propTypes = {
+  name: PropTypes.string.isRequired,
+  className: PropTypes.string,
+  location: PropTypes.shape({
```
ifsimicoded: should we define some location props as required?

src/js/modules/Contact/Address.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import MapMarkerSVG from 'js/svg/MapMarker';
+import PropTypes from 'prop-types';
+
+const Address = ({ className, name, location }) => (
+  <div className={`${className} coa-ContactAddress`}>
```
ifsimicoded: The more I look at this, I see the problem in how I structured contact items. Instead of passing a class through (which couples this component to a parent) we should just statically class all of the contact components with 'coa-ContactDetail' or 'coa-ContactItem' or something so the individual contact components can have standalone styles.

src/js/modules/Contact/Phone.js
```diff
@@ -12,4 +12,10 @@ const ContactPhone = ({ className, phone, ttyphone }) => (
   </div>
 );
 
-export default ContactPhone;
+Phone.propTypes = {
+  phone: PropTypes.string.isRequired,
+  className: PropTypes.string,
+  ttyphone: PropTypes.string,
```
ifsimicoded: if ttyphone is not required, we should probably update the JSX to optionally not render the property if it's not set

src/js/modules/HtmlFromAdmin.js
```diff
@@ -1,18 +1,16 @@
...
   <div className="coa-HtmlFromAdmin">
-    <h2 className="coa-HtmlFromAdmin__title">{intl.formatMessage(i18nMessages.htmlFromAdminSectionTitle)}</h2>
+    <h2 className="coa-HtmlFromAdmin__title">{title}</h2>
     <div className="coa-HtmlFromAdmin__content" dangerouslySetInnerHTML={{__html: content}} />
   </div>
 );
 
-export default injectIntl(HtmlFromAdmin);
+HtmlFromAdmin.contextTypes = {
+  title: PropTypes.string,
```
ifsimicoded: Same, if title isn't required, we should add a conditional to render it.

src/js/page_sections/Header.js
```diff
@@ -73,11 +75,15 @@ class Header extends Component {
             </div>
           </div>
         </div>
-        <Menu isMenuOpen={this.state.menuIsOpen} toggleMenu={this.toggleMenu} />
+        <Menu
+          isMenuOpen={this.state.menuIsOpen}
+          toggleMenu={this.toggleMenu}
+          navigation={this.props.navigation[this.props.intl.locale]}
```
ifsimicoded: I think props is already destructured on line 44 -- you could add navigation to that destructure statement and shorten line 81 to `navigation={navigation[intl.locale]}`

src/stories/PageBanner.stories.js
```diff
@@ -0,0 +1,20 @@
...
+import { checkA11y } from '@storybook/addon-a11y';
+import { IntlProvider } from 'react-intl';
+
+
+import PageBanner from 'js/modules/PageBanner';
+
+import dummyImage from 'stories/images/pagebanner_test.jpg';
+
+storiesOf('PageBanner', module)
+  .addDecorator(checkA11y)
```
ifsimicoded: Since we're adding checkA11y and IntlProvider on all components, I think it makes sense to do this globally once. 
You can also add a decorator globally for all stories like this in .storybook/config.js according to the docs
```
import React from 'react';
import { configure, addDecorator } from '@storybook/react';
import 'css/coa.css'

const SomeDecorator = (storyFn) => (
  <div style={{
    border: '10px solid burlywood'
  }}>
    { storyFn() }
  </div>
);

addDecorator(SomeDecorator);

// automatically import all files ending in *.stories.js
const req = require.context('../src/stories', true, /.stories.js$/);
function loadStories() {
  req.keys().forEach((filename) => req(filename));
}

configure(loadStories, module);

```

src/js/page_sections/Header.js
```diff
@@ -1,5 +1,7 @@
 import React, { Component } from 'react';
 import { defineMessages, injectIntl } from 'react-intl';
+import { withSiteData } from 'react-static';
```
ifsimicoded: I can't remember how we landed on moving this up to Header. But I'm thinking all the siteData 🎶 and 💃  should be done in LanguageWrapper and passed into Header (or whatever sub component may need it). This way, if you want to make a storybook out of header too.

src/js/page_sections/Menu/Menu.js
```diff
@@ -192,4 +191,10 @@ const MobileFooter = injectIntl(({intl}) => (
   </div>
 ))
 
-export default withSiteData(injectIntl(Menu));
+Menu.childContextTypes = {
```
ifsimicoded: Did you mean PropTypes instead of childContextType? If not, can you explain how're you're using this?

ifsimicoded: Also I like your destructuring approach at the top of the render function so it's clear what props are being used. With propTypes, it's probably redundant, but I like extra obvious things. could we do that in Menu.js as well. So
```
const { menuIsOpen, toggleMenu, navigation } = this.props;
```

src/js/page_sections/Menu/Menu.js
```diff
@@ -81,7 +80,7 @@ class Menu extends Component {
   }
 
   render() {
-    const { allThemes } = this.props.navigation[this.props.intl.locale];
+    const { allThemes } = this.props.navigation;
 
     return allThemes.edges.length && (
```
ifsimicoded: Unrelated to this PR, but could we make a cleanNavigation helper function, like those in js/helpers/cleanData.js so Menu.js isn't expecting data in a graphql specific format?
So instead of having to loop through things with edges and nodes could we just loop through a regular collection of objects? No need to nest it in allThemes either.

src/js/pages/Theme.js
```diff
@@ -19,7 +19,8 @@ const Theme = ({ theme }) => {
       </div>
 
       <div className="wrapper container-fluid">
-        <TileGroupSet groups={topics}
+        <TileGroupSet
+          groups={topics}
           tileKey={'services'}
           groupTitleSubPath={'topics'}
```
ifsimicoded: The more I see this subpath thing everywhere for topic pages and theme pages, I get bad vibes. Maybe there's a better way to define topic links and theme links... or links in general... in just one place. This type of thing seems like it's a middleware transformation that should be done once wherever the topmost data comes through. Maybe in a clean data step at the page level with all the other clean data steps for RouteData. Or in LanguageWrapper for SiteData things. What do you think?

src/stories/Footer.stories.js
```diff
@@ -5,7 +5,7 @@ import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
 import { linkTo } from '@storybook/addon-links';
 import { checkA11y } from '@storybook/addon-a11y';
-import { Route, Link, MemoryRouter } from 'react-router-dom';
+import { MemoryRouter } from 'react-router-dom';
```
ifsimicoded: Can you tell me more about what MemoryRouter is used for in storybook? I'm taking a brief break from reading documentation :)

src/js/modules/HtmlFromAdmin.js
```diff
@@ -1,18 +1,16 @@
...
+  content: PropTypes.string.isRequired,
```
ifsimicoded: Not 100% sure why -- but I get a weird warning that says content is required but not set -- tho it's def set.
screen shot 2018-04-09 at 8 41 11 pm
screen shot 2018-04-09 at 8 41 43 pm

src/stories/Menu.stories.js
```diff
@@ -38,10 +40,10 @@ storiesOf('Menu', module)
       <ul className="coa-Menu__list">
```
ifsimicoded: Does this mean that MenuList should be a component? and not MenuItem? MenuItem clearly isn't stand-alone. But it seems like the list is? Not sure what makes sense here.

src/stories/PageBreadcrumbs.stories.js
```diff
@@ -0,0 +1,38 @@
...
+import { MemoryRouter } from 'react-router-dom';
+
+
+
+import PageBreadcrumbs from 'js/modules/PageBreadcrumbs';
+
+const grandparent = {
+  slug: "grandparent",
+  text: "Grandparent Theme",
+  subpath: "themes"
```
ifsimicoded: so many subpaths is bad joo joo. Or maybe I'm just hungry. really hard to tell the difference.

src/stories/Recollect.stories.js
```diff
@@ -0,0 +1,22 @@
...
+
+
+import Recollect from 'js/modules/Recollect';
+
+
+storiesOf('Recollect', module)
+  .addDecorator(checkA11y)
+  .add('recollect', () => (
+    <IntlProvider locale="en">
+      <div className="wrapper container-fluid">
```
ifsimicoded: Would it just make sense to have a global wrapper container-fluid decorator? or maybe just a decorator method we can use for this as needed.

src/stories/Recollect.stories.js
```diff
@@ -0,0 +1,22 @@
...
+        <code style={{ color: 'white', backgroundColor: '#170821', padding: '10px', display: 'block' }}>
+          <p>Storybook puts everything inside an iframe. Recollect says "This widget cannot function properly in an iframe. Please contact support@recollect.net for help with embedding the widget." 🙃 </p>
```
ifsimicoded: Oh man, third party stuff is just 💩

src/stories/SecondaryContentBanner.stories.js
```diff
@@ -0,0 +1,30 @@
...
+storiesOf('SecondaryContentBanner', module)
+  .addDecorator(checkA11y)
+  .addDecorator(story => (
+    <MemoryRouter initialEntries={['/']}>{story()}</MemoryRouter>
+  ))
+  .add('SecondaryContentBanner', () => (
+    <IntlProvider locale="en">
+      <SecondaryContentBanner>
+        <p>
+          <FormattedMessage
```
ifsimicoded: Maybe we can put dummy text in, instead of FormattedMessage react-intl component. It's not required to test the component.

src/stories/Steps.stories.js
ifsimicoded: This makes a lot of sense. It's my understanding that this was the simplest way for wagtail to send janis data. so the hope is we'll just be looping at some point over raw content and not dumping mark up. But maybe until that change happens, it makes a lot more sense to rename this to be really clear. Maybe even a comment in the component that shows the expected data format.

ifsimicoded: stepsHtmlFromAdmin?

src/stories/ThreeOneOne.stories.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import { storiesOf } from '@storybook/react';
+import { checkA11y } from '@storybook/addon-a11y';
+import { IntlProvider } from 'react-intl';
+
+
+import ThreeOneOne from 'js/page_sections/ThreeOneOne';
+import jsonFileData from '__tmpdata/pages';
```
ifsimicoded: should we keep static data for stories in one place? Maybe we can copy this file to stories/static_data/ so it can work for both until we get 311 data in joplin for the static app.

src/stories/HeroHome.stories.js
```diff
@@ -1,27 +1,21 @@
...
-
 import { storiesOf } from '@storybook/react';
-import { action } from '@storybook/addon-actions';
-import { linkTo } from '@storybook/addon-links';
 import { checkA11y } from '@storybook/addon-a11y';
-import { Route, Link, MemoryRouter } from 'react-router-dom';
+import { MemoryRouter } from 'react-router-dom';
 
-import HeroHome from 'js/modules/HeroHome';
+import dummyImage from 'stories/images/herohome_test.jpg';
```
ifsimicoded: can images live in static_data too?

src/js/pages/Theme.js
ifsimicoded: We already have methods that build out service links in cleanData.js (cleanServiceLinks, cleanRelatedServiceLinks) ... maybe we can just add more helper methods?

src/stories/static_data/tileData.js
```diff
@@ -0,0 +1,102 @@
...
+  {
+    "url": "/services/trash-recycling-pickup",
+    "text": "Look up your trash, recycling, and compost pickup days"
+  }
+]
+
+export const tileGroupData = {
+  "edges": [
+    {
+      "node": {
```
ifsimicoded: Mentioned this elsewhere, but I think storybook yet again shows us how we've tightly coupled our components -- this time to graphQL style data.

src/js/pages/Services.js
```diff
@@ -30,7 +30,7 @@ const Services = ({ allServices, intl }) => {
   return (
     <div>
       <div className="wrapper wrapper--sm container-fluid">
-        <PageHeader
+        <PageHeader
```
benweatherman: Bold move to add a space here

src/js/modules/Contact/Address.js
mateoclarke: Not sure what you mean when you say "statically class" but in general this sounds good. I'll add this as technical debt to the backlog.

mateoclarke: going to give what I think you're saying a shot.

src/stories/PageBanner.stories.js
mateoclarke: 👑

src/stories/Footer.stories.js
mateoclarke: I got this error `You should not use <Route> or withRouter() outside a <Router>` when I was adding a story so I googled it and MemoryRouter was the answer. 

From React Router 4 docs:
> `<MemoryRouter>` is a `<Router>` that keeps the history of your "URL" in memory (does not read or write to the address bar). Useful in tests and non-browser environments like React Native.

src/js/modules/HtmlFromAdmin.js
mateoclarke: should've `.propTypes`, not `.contextTypes`

src/stories/Menu.stories.js
mateoclarke: Just needed that class so styling would work. MenuList could be a component, but its not yelling out at me to be moved into its own component... yet

src/stories/Recollect.stories.js
mateoclarke: thought about this but we don't want it for Header, Footer, PageBanner, LanguageSelect and maybe a few others.

PR #215, opened 2018-04-05
Feedback Form

ifsimicoded:
So primary stuff below BUT, I also updated some header stuff for alignment since I wanted a break from the form. I'm a horrible person for not creating a new branch. The header changes live in one commit -- https://github.com/cityofaustin/janis/pull/215/commits/a120938faa99b0f0885d12af6bd7c6e80bb6d781

addresses https://github.com/cityofaustin/techstack/issues/182
Completed:
- [x] form redesign implemented
- [x] data logged in google analytics for each step
- [x] data sent to gh issue -- https://github.com/cityofaustin/alpha-feedback
- [x] poorly handled error messaging
- [x] clean up unused sass from previous form component

To be completed:
- [ ] storybook component -- since this leans on an environment variable (I'll wait for @mateoclarke's changes to better decouple our components from environment vars.
feedback

@cthibaultatx the transition from emoji's to the second step and then to success step is a bit jarring -- any ideas?
 Or maybe i'm being extra sensitive.


src/js/modules/FormFeedback.js
```diff
@@ -0,0 +1,231 @@
...
+    this.state = {
+      step: 1,
+      loading: false,
+      emoji: null,
+      feedback: null,
+      error: null
+    };
+  }
+
+  logEvent(action, label) {
```
mateoclarke: I'm confused by this. It seems like you are defining `logEvent` as a function that calls the `logEvent` function you imported from helpers. Maybe just using different naming would help? Also, why not use the arrow function syntax like you did below?

src/js/modules/FormFeedback.js
```diff
@@ -0,0 +1,231 @@
...
+    logEvent({
+     category: 'FORM__site-feedback',
+     action: `${action}__step-${this.state.step}`,
+     label: label
+    });
+  }
+
+  handleEmojiClick = (e) => {
+    this.logEvent('emoji-click', JSON.stringify({
+        'emoji':e.currentTarget.value
```
mateoclarke: space after `:`?

src/js/modules/FormFeedback.js
```diff
@@ -0,0 +1,231 @@
...
+  logEvent(action, label) {
+    logEvent({
+     category: 'FORM__site-feedback',
+     action: `${action}__step-${this.state.step}`,
+     label: label
+    });
+  }
+
+  handleEmojiClick = (e) => {
+    this.logEvent('emoji-click', JSON.stringify({
```
mateoclarke: does GA expect hyphens not camel case? I guess that doesn't matter just wondering.

src/js/modules/FormFeedback.js
```diff
@@ -0,0 +1,231 @@
...
+  handleTextAreaChange = (e) => {
+    this.setState({
+      feedback: e.currentTarget.value
+    });
+  }
+
+  handleSubmit = (e) => {
+    e.preventDefault();
+
+    this.logEvent('send-feedback-click', JSON.stringify({
```
mateoclarke: Why do we have to stringify this object as we pass it as an argument? Would it be better for that stringify-ication to happen inside the function if we're going to have to do it all the time when its invoked?

src/js/modules/FormFeedback.js
```diff
@@ -0,0 +1,231 @@
...
+            >
+              <img src={emojiNeutral} alt={intl.formatMessage(i18nMessages.emojiNeutral)} />
+            </button>
+            <button onClick={this.handleEmojiClick}
+              value="1"
+              aria-label={intl.formatMessage(i18nMessages.emojiSlightlySmiling)}
+            >
+              <img src={emojiSlightlySmiling} alt={intl.formatMessage(i18nMessages.emojiSlightlySmiling)} />
+            </button>
+            <button onClick={this.handleEmojiClick}
```
mateoclarke: Lots of repetition around this button. Wondering if thats a clue that we should make an `<EmojiButton />` stateless functional component to pass an emoji object as a prop. Where that object has keys like `img`, `altText`, and `value`. Doesn't need its own file imo.

src/js/modules/FormFeedback.js
```diff
@@ -0,0 +1,231 @@
...
+          <form>
+            { this.state.error && (
+              <p className="coa-FormFeedback__error">Oh no, something went wrong! Please, try submitting your feedback again.</p>
+            )}
+            <label htmlFor="site-feedback-textarea"
+              className="coa-sr-only"
+            >{intl.formatMessage(i18nMessages.step2titleb)}</label>
+            <textarea id="site-feedback-textarea"
+              onChange={this.handleTextAreaChange}></textarea>
+            <p>Please do not enter private information because your feedback will be public.</p>
```
mateoclarke: I know this is a content thing, but this struck me as odd. If I saw this messaging, I would hesitate to give feedback. Have you talked to them about that?

src/js/modules/FormFeedback.js
```diff
@@ -0,0 +1,231 @@
...
+import { logEvent } from 'js/helpers/googleAnalytics';
+import SectionHeaderSerif from 'js/modules/SectionHeaderSerif';
+
+import emojiDisappointed from 'images/emojis/disappointed.png';
+import emojiSad from 'images/emojis/sad.png';
+import emojiNeutral from 'images/emojis/neutral.png';
+import emojiSlightlySmiling from 'images/emojis/slightly_smiling.png';
+import emojiGrinning from 'images/emojis/grinning.png';
+
+const i18nMessages = defineMessages({
```
mateoclarke: I know we've gotten into a patter of including these defineMessage object literals inside the component. Just thinking about whether they should be imported from another i18nMessage dir or something instead? Could that help with shareability between components? No real constructive comment, just thinking out loud.

src/css/helpers/_helpers.scss
```diff
@@ -1,32 +1,10 @@
-.nowrap {
-  white-space: nowrap;
+.coa-text-spacer--vertical:after {
+  border-left: 1px solid $coa-color-purple-dark;
```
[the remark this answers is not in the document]
ifsimicoded: I think it behaves as you would expect -- this is the spacer between the airport and 311 link and the Menu button and the alpha.austin.gov link.

screen shot 2018-04-06 at 11 59 02 am

src/js/modules/SectionHeaderSerif.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SectionHeaderSerif = ({ title }) => (
```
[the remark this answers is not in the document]
ifsimicoded: Yea, I noodled on this. I thought the sass was sufficiently different (since the section header is sized differently, has an underline thing and has an arrow option), that I made a new component so it wouldn't seem like the two were related. 
screen shot 2018-04-06 at 12 00 51 pm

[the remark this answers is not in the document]
ifsimicoded: Open to mashing them together tho -- let me think on this. I can make the underline an option as well.

src/js/modules/FormFeedback.js
ifsimicoded: Yea, renaming might help -- basically I'm using the component scoped logEvent to call the more generic helper logEvent with some automatic behavior baked in -- passing the step and form and form name. I can make this better for sure.

I didn't make it an arrow function because it's not an event handler so I don't need to create a this scope for it.

src/js/modules/FormFeedback.js
ifsimicoded: ✅

src/js/modules/FormFeedback.js
mateoclarke: So would it be `emoji__click`? Why BEM if it doesn't relate to CSS?

ifsimicoded: emoji-click cause BEM hyphenates compounds/phrases. Honestly, it can be anything. But with my GA experience, any naming convention is necessary for simplifying analysis later. Some of the biggest issues I've seen making GA data usable is that 20 different people had 20 different ideas as to how to name things, so aggregation was impossible. I chose BEM because we're already using it mostly, and because we're trying to name things related to html elements.

src/js/modules/SectionHeaderSerif.js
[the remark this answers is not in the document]
ifsimicoded: Updated! Thanks for the feedback on this.

src/js/modules/FormFeedback.js
ifsimicoded: Implemented a generic logFormEvent in our googleAnalytic.js helper functions. the class method logEvent now calls the helper method logFormEvent. Hopefully this is clearer .

src/js/modules/FormFeedback.js
ifsimicoded: Added a logFormEvent which JSON.stringifies in one place.

src/js/modules/FormFeedback.js
ifsimicoded: Updated this to be a map function over emoji keys. Hopefully this handles the repetition well enough

cthibaultatx:
@ifsimicoded adding the updated comp 
screen shot 2018-04-11 at 12 31 57 pm

PR #220, opened 2018-04-16
Sd techdebt svgmodules

ifsimicoded:
This was from some great PR feedback @mateoclarke provided.
- Update all svgs to be DRY-er
- Set svg sizes responsively based on parent font size if not otherwise defined
- make magic numbers variables


src/css/page_sections/Menu/_Menu.scss
```diff
@@ -43,15 +43,11 @@
 }
 
 .coa-Menu__close-btn {
-  background: transparent;
```
ifsimicoded: these style overrides are already in place for buttons in uswds overrides

src/js/svg/Svg.js
```diff
@@ -7,4 +8,8 @@ const SVG = ({ title, children, ...rest }) => (
   </svg>
 );
 
+SVG.propTypes = {
+  children: PropTypes.element.isRequired
+};
+
 export default SVG;
```
mateoclarke: dig it

ifsimicoded:
Gotcha on the vertical alignment issue -- fixed.

PR #222, opened 2018-04-16
BrowserStack Audit

mateoclarke:
_Closes https://github.com/cityofaustin/techstack/issues/146_
_Closes https://github.com/cityofaustin/coa-dev-internal/issues/21_
_Starts https://github.com/cityofaustin/techstack/issues/308_

- Update the README
- Add a section about browser compatibility
- Polyfills for React Static in IE & Intl in IE and Safari < 10

There is definitely more comprehensive work we could be doing but I'm hesitant to over commit us this early in the project and without a dedicated QA resource. I really like this guideline cfpb has: https://github.com/cfpb/development/blob/master/guides/browser-support.md

Logged some more issues as a result of browsing that can be addressed in another PR.
- https://github.com/cityofaustin/techstack/issues/308
- https://github.com/cityofaustin/techstack/issues/307


README.md
```diff
@@ -1,15 +1,136 @@
...
 Your react app should be running at http://localhost:3000/
 
-## Adding dependencies via yarn
+**As a static build**
+To run the site locally as a static build, the way it works in production, see the [Static build scripts](#static-build-scripts) section below.
+
+
+## Project Info
+
+This work began under the he Austin Digital Services Discovery Project. Design, Technology, and Innovation Fellows worked in partnership between two City departments, Communications Public Information and Communications and Technology Management.
```
benweatherman: `under the he Austin`

README.md
```diff
@@ -1,15 +1,136 @@
...
+
+For more information about the concepts and architectural decisions guiding this work, take a break from this README and check out some of what we have written on Medium and our project page:
+
+- https://medium.com/civiqueso/open-source-city-cms-part-1-607a58b32356
+- http://projects.austintexas.io/projects/austin-digital-services-discovery/about/Dev/
+- http://projects.austintexas.io/projects/austin-digital-services-discovery/our-technical-approach/open-by-default/
+
+### React-Static
+
+This project uses [React-Static](https://github.com/nozzle/react-static) as a base framework for building static-progressive a React application/website. It's designed to with considerations for SEO, site performance, and user/developer experience.
```
benweatherman: `for building static-progressive a React application/website`

`It's designed to with considerations`

README.md
```diff
@@ -1,15 +1,136 @@
...
+
+TODO: more on this decision + evaluation
```
benweatherman: meh

README.md
```diff
@@ -34,17 +155,23 @@ docker exec --interactive --tty janis yarn run build-langs
 ```
 
 Note: if you're running the docker container built by serve-build.sh you'll have to update the container name from `janis` to `janis-build` in the above command.
-___
 
+### Static build scripts
 
-## Design assets
+Since we use React-Static as our framework to render our React components as a static progressive website, its important for us to be able to test the final static build locally. In order to do this, we have two scripts.
```
benweatherman: `its important`

README.md
```diff
@@ -34,17 +152,23 @@ docker exec --interactive --tty janis yarn run build-langs
...
 
-Current design mockups are stored in Google Drive as Sketch Files:
-https://drive.google.com/drive/u/1/folders/1Xg6739ixhIDrb7MdJAnO4_zCRu3kDmZQ
+**To build**
+```
+./scripts/build.sh
+```
 
----
+**To build and server**
```
ifsimicoded: **To build and serve**  - r

ifsimicoded:
Do we want to invest in making our serve-local build operable in IE so we can live debug in IE?
The app seems to die on the strip-ansi package that CRA has a fixed issue logged.
https://github.com/facebook/create-react-app/issues/2691

README.md
mateoclarke: good point, we made the build script before the build and serve and I haven't used it since.

mateoclarke:
@ifsimicoded thanks for detail around how we would make serve-local work in IE. If its as easy as downgrading a dependency, I'd say why not give it a try and log it as an issue. But when it comes to live debugging in IE, wouldn't it be better to use `serve-build` since thats what a user would experience in production? But more is better, right?

PR #223, opened 2018-04-17
refactor tile component layout logic. clean data in toplevel components

ifsimicoded:
- I refactored the TileGroupSet components to lean more on sibling counts via CSS selectors and viewport widths so our layout decisions could live outside of our JS and in our CSS. 
- I also removed the data cleaning logic out of the TileGroup component so the component does not need to be aware of graphQL nodes and edges. (addresses some of https://github.com/cityofaustin/techstack/issues/244)
- This also handles "body inside tile group should be centered" from https://github.com/cityofaustin/techstack/issues/236 

* Note this PR merges in to another branch with an open PR. I separated the two, since this refactor got heavy.

### DIFF
--- a/src/css/modules/_TileGroupSet.scss
+++ b/src/css/modules/_TileGroupSet.scss
@@ -1,4 +1,45 @@
 .coa-TileGroupSet {
   display: flex;
   justify-content: center;
+
+  .coa-TileGroup {
+    margin: 1%;
+    width: 100% - 1% * 2;
+  }
+
+  @include media($coa-medium-screen) {
+
+    /* two or more groups */
+    .coa-TileGroup:nth-last-child(n+2):first-child,
+    .coa-TileGroup:nth-last-child(n+2):first-child  ~ .coa-TileGroup {
+      width: 50% - 1% * 2;
+      .coa-TileGroup__tile {
+        flex-basis: 100%;
+        max-width: 100%;
+      }
+    }
+  }
+
+  @include media($coa-large-screen) {
+
+    /* two groups */
+    .coa-TileGroup:nth-last-child(2):first-child,
+    .coa-TileGroup:nth-last-child(2):first-child  ~ .coa-TileGroup {
+      width: 50% - 1% * 2;
+      .coa-TileGroup__tile {
+        flex-basis: 50%;
+        max-width: 50%;
+      }
+    }
+
+    /* three or more groups */
+    .coa-TileGroup:nth-last-child(n+3):first-child,
+    .coa-TileGroup:nth-last-child(n+3):first-child ~ .coa-TileGroup {
+      width: 33% - 1% * 2;
+      .coa-TileGroup__tile {
+        flex-basis: 100%;
+        max-width: 100%;
+      }
+    }
+  }
 }
--- a/src/js/helpers/cleanData.js
+++ b/src/js/helpers/cleanData.js
@@ -55,14 +55,16 @@ export const cleanRelatedServiceLinks = (links) => {
   });
 };
 
-export const cleanServiceLinks = (links) => {
+export const cleanLinks = (links, pathPrefix) => {
 
   if(!links || !links.edges) return null;
 
   return links.edges.map(({node: link}) => {
+    const {title, slug, ...rest} = link;
     return {
-      url: `/services/${link.slug}`,
-      text: link.title
+      url: `${pathPrefix || ''}/${slug}`,
+      text: title,
+      ...rest
     }
   });
 };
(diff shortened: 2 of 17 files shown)


src/js/modules/SectionHeader.js
```diff
@@ -1,26 +1,18 @@
...
 
-
-const SectionHeader = ({ isSerif, hasHighlight, hasArrow, children }) => (
+const SectionHeader = ({ isSerif, hasHighlight, children }) => (
   <h2 className={`coa_SectionHeader
     ${hasHighlight ? 'coa_SectionHeader--hasHighlight' : ''}
     ${isSerif ? 'coa_SectionHeader--isSerif' : ''} `
   }>
     {children}
-    {hasArrow && (
```
ifsimicoded: I moved the arrow svg out, so I could nest it within a link that's passed in as children.

src/css/modules/_TileGroupSet.scss
```diff
@@ -1,4 +1,45 @@
...
+      width: 50% - 1% * 2;
+      .coa-TileGroup__tile {
+        flex-basis: 50%;
+        max-width: 50%;
+      }
+    }
+
+    /* three or more groups */
+    .coa-TileGroup:nth-last-child(n+3):first-child,
+    .coa-TileGroup:nth-last-child(n+3):first-child ~ .coa-TileGroup {
```
mateoclarke: 🔮  you're a wizard Harry

src/js/helpers/cleanData.js
```diff
@@ -55,14 +55,16 @@ export const cleanRelatedServiceLinks = (links) => {
   });
 };
 
-export const cleanServiceLinks = (links) => {
+export const cleanLinks = (links, pathPrefix) => {
```
mateoclarke: haha, I thought you didn't like pathPrefix.

src/js/helpers/cleanData.js
```diff
@@ -55,14 +55,16 @@ export const cleanRelatedServiceLinks = (links) => {
...
   if(!links || !links.edges) return null;
 
   return links.edges.map(({node: link}) => {
+    const {title, slug, ...rest} = link;
     return {
-      url: `/services/${link.slug}`,
-      text: link.title
+      url: `${pathPrefix || ''}/${slug}`,
+      text: title,
+      ...rest
```
mateoclarke: `...rest` is pretty great.

src/js/modules/TileGroupSet.js
```diff
@@ -3,47 +3,34 @@ import PropTypes from 'prop-types';
...
-            direction={shouldTilesStackVertical ? 'column' : false}
-          />
-        )
-      })
-    }
+    <div className="coa-TileGroupSet">
+      <div className="row">
+      {
+        tileGroups.map(({ tiles, text, url, description }, index) => {
+          if (!tiles.length) return null;
```
mateoclarke: This is a n00b question. Why is returning `null` better than returning `false`?

src/js/pages/Home.js
```diff
@@ -30,7 +30,8 @@ const i18nMessages = defineMessages({
 });
 
 const Home = ({ topServices, image, intl }) => {
-  const serviceLinks = cleanServiceLinks(topServices);
+  //TODO: clean data where sourced
```
mateoclarke: I don't understand the TODO

mateoclarke: oh, you mean we should expect Joplin to give us the full path in the query? If so, we should log this so Ben (or whoever) knows we want the query changed.

src/js/helpers/cleanData.js
ifsimicoded: I don't like it showing up in multiple places since it's more prone to errors or being missed if we update the path. If we build the links in one place, it's cool.

src/js/helpers/cleanData.js
ifsimicoded: ...dope

src/js/modules/TileGroupSet.js
ifsimicoded: mmmmm, this is just a personal preference. I think of null as an empty type. Where as false is its own data type -- boolean.  so whenever my functions are expected to return a particular data type (in this case JSX), if it can't return that type, I like to return an empty vs a whole other data type.

src/js/pages/Home.js
ifsimicoded: Oh no, I'm meaning that data cleaning should happen in one place -- it could be Joplin, or maybe it happens in our react.static.config file. OR some other middleware layer for data transformations. But basically we should reformat data in one spot so when data formatting needs change, we only change it in one place AND so different components which render the same data in different place show that data consistently (or consume the same transformed data). But I'm not sure where this cleaning happens. for now, transformation consistency is taken care of since one shared method is performing the transform.

PR #229, opened 2018-04-26
Feedback Form Url and User Agent Logging

ifsimicoded:
closes https://github.com/cityofaustin/techstack/issues/182

feedback form logs user's current url and device info as shown here:
https://github.com/cityofaustin/alpha-feedback/issues/18


src/components/FormFeedback/index.js
```diff
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import { defineMessages, injectIntl } from 'react-intl';
+import parser from 'ua-parser-js';
```
mateoclarke: could this variable be more descriptive? like `userAgent`

PR #233, opened 2018-04-30
upgrade react static

ifsimicoded:
Homepage load [issue](https://github.com/cityofaustin/techstack/issues/289) fixed in latest react static -- so let's upgrade. Initially, we downgraded to fix the issue introduced in react-static ~5.7
https://github.com/nozzle/react-static/issues/567

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "react-load-script": "^0.0.6",
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
-    "react-static": "^5.6.8",
+    "react-static": "^5.8.5",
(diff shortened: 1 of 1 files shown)


mateoclarke:
sorry if this is a dumb question...

You updated the package.json but I was also expecting to see a yarn.lock file change reflecting the upgrade. Do we need to run yarn on a docker instance to get that file updated too?

if not, 🚢

ifsimicoded:
GOOD CALL! I did run a docker instance, but maybe I never committed the yarn.lock. Not sure where that got lost, but I should've committed it as well. 🙏 Thanks for saving lives.

PR #234, opened 2018-05-01
Static storybook

mateoclarke:
We played with a few ways to do this before landing on something simple...

[These docs](https://storybook.js.org/basics/exporting-storybook/) gave us the info we needed to build a static export of storybook and update our package.json.

[This plugin](https://github.com/storybooks/storybook-deployer) gave us a script that would automatically deploy the output static file to Github Pages via a `gh-pages` branch. But this had issues because it couldn't run in Docker (our docker container doesn't have git installed). So we would have to run a build script in Docker with a "volume" reference to copy the output file locally. Then we would run the deploy script outside of Docker and point it at the created output file.

This was doomed to get out of date quickly and @benweatherman said that letting a 3rd party script hijack your git is sketchy. Then @ifsimicoded called me a cowboy. 🤠 

So, we decided to let Docker and nginx handle the serving of the static storybook folder within our existing static build command. This way, for whatever instance of `xyz.austin.gov`, there will be a component library whose code is in sync with janis at `xyz.austin.gov/storybook/`. Simi and Ben said that is so v cool. 😎 

... still need to figure out that trailing slash but there is an issue for that. 
https://github.com/cityofaustin/techstack/issues/342

_Closes https://github.com/cityofaustin/techstack/issues/266_

screen shot 2018-05-01 at 6 57 24 pm


src/stories/static_data/threeoneoneData.js
```diff
@@ -1,64 +1,131 @@
 const threeoneoneData = [
   {
-    text: 'Report Found Animal your Holding',
     url:
-      'http://311.austintexas.gov/reports/new?service_id=535ffd34844eedb146000019',
+      'http://311.austintexas.gov/reports/new?service_id=5579f3bd47199e0be40000dd',
```
ifsimicoded: Thanks for updating these!!!!

Dockerfile.build
```diff
@@ -15,6 +15,7 @@ RUN yarn
 COPY intl.buildlangs.js /app/intl.buildlangs.js
 COPY static.config.js /app/static.config.js
 COPY .babelrc /app/.babelrc
+COPY .storybook /app/.storybook
```
ifsimicoded: Can we update our serve-local process, too, to serve storybook?
And then can we delete the old Dockerfile.storybook and serve-storybook files?

ifsimicoded: Maybe we can get storybook to serve at a different port than janis, and then we need to just let docker know to publish that port?

ifsimicoded:
giphy

PR #235, opened 2018-05-03
Foster Pet Form - React form library exploration

mateoclarke:
This will probably only be an exploratory PR using a couple different libraries for forms:
  - [react-jsonschema-form](https://github.com/mozilla-services/react-jsonschema-form)
    - really like the separation of schemas from code markup.
    - library covers a lot, bunch of documentation but hard at times to comb through to find what you're looking for.
    - disappointed that conditionals are not better supported out of the box. Pretty manual pattern for implementation that has you save the schema to state and make manipulations in the Form's `onChange` event: https://jsfiddle.net/cowbellerina/zbfh96b1/
    - even though a lot is manual, it's nice that you can create your own ui widgets as separate re-usable React components
    - ~validating `onBlur` seems tricky~ nvm, not tricky: https://github.com/mozilla-services/react-jsonschema-form/issues/617
  - [formik](https://github.com/jaredpalmer/formik)
    - way different approach using the HOC pattern to give special functionality to a manually coded HTML form. helps with:
         1) Transforming props to form state: `mapPropsToValues` 
        2) Validation and error messages dependent on Yup library `validationSchema`
          3) Handling form submission. `handleSubmit`
    - doesn't seem mutually exclusive with `react-jsonschema-form`. We could wrap our jsonschema-form in a formik HOC, but that seems like overkill, at least right now.
    - had dependency installation issues with `yup` around 'babel-runtime/helpers'. Maybe this isn't worth troubleshooting.
  - [Redux Form](https://redux-form.com/7.3.0/docs/gettingstarted.md/)
    - this is a popular solution for managing state but its dependent on Redux as a dependency. Redux can be really nice but it has a complicated architecture for new people and a high learning curve. It doesn't seem like the right thing to add to this project at this point in time. Formik sells itself as a competitor to Redux Forms without the Redux dependency.

TL;DR:
- After reading docs and experimenting with this for a couple days too long, I feel ok about moving forward with `react-jsonschema-form` and seeing how far that gets us. I'm going to reach out to some VA folks and see if they have example forms open-sourced I could examine and see if they've figured out something to deal with `onBlur` events. This will just be one piece of a larger puzzle about how custom forms work. On to the next blocker...
- going to branch from here and close this PR.


mateoclarke:
Closing this out in favor of work being done in #241

mateoclarke:
Closes https://github.com/cityofaustin/techstack/issues/370

PR #237, opened 2018-05-08
more i18n cleanup

ifsimicoded:
I built this branch off of PR 236 to try to keep the changes more manageable.
- make I18n rtl/ltr language direction decorator so we can use it as a component decorator in storybook
- consolidate i18n components in i18n folder
- add skip to nav component and translation definition
- prettier automated formatting fixes


src/components/Pages/Home.js
```diff
@@ -5,7 +5,7 @@ import { injectIntl } from 'react-intl';
 import { misc as i18n2, services as i18n3 } from 'js/i18n/definitions';
 
 import WorkInProgress from 'components/WorkInProgress';
-import PageSubBanner from 'components/PageSubBanner';
+import PageNotificationBanner from 'components/PageNotificationBanner';
```
ifsimicoded: Talked to @cthibaultatx  about the rename and we landed on something between banner and what we originally labelled (PageNotification) in our pin up.

.storybook/config.js
```diff
@@ -5,23 +5,29 @@ import { checkA11y } from '@storybook/addon-a11y';
...
 setIntlConfig({
-    locales: SUPPORTED_LANG_CODES,
-    defaultLocale: DEFAULT_LANG,
-    getMessages
+  locales: SUPPORTED_LANG_CODES,
+  defaultLocale: DEFAULT_LANG,
+  getMessages,
 });
 
+addDecorator(I18nClassDecorator);
```
mateoclarke: 🌍

src/components/I18n/I18nController.js
```diff
@@ -9,19 +9,19 @@ import { find } from 'lodash';
...
 
-// page_sections
+import I18nDecorator from 'components/I18n/I18nDecorator';
+import SkipToMain from 'components/PageSections/SkipToMain';
 import LanguageSelectBar from 'components/PageSections/LanguageSelectBar';
 import Header from 'components/PageSections/Header';
 import Footer from 'components/PageSections/Footer';
 
-class LanguageWrapper extends Component {
+class I18nController extends Component {
```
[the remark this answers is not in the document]
ifsimicoded: Thanks for sharing the article! My thinking was what Controller means in a broad sense -- code that acts as a processing layer between data and views. So this I18nController's ONLY responsibility is to determine what language/locale we need to set for the user and to pass the appropriate data to view modules based on that language. So this component shouldn't have anything to do with layout or styles, etc. 

I was even thinking of moving out the children components to a general page layout component (or throwing them in App.js as children), in case we have different Header, Footer, Etc, needs for other city pages and so this controller isn't responsible for page layout.

Hopefully that makes some sense.

PR #238, opened 2018-05-10
translation workflow how-to

ifsimicoded:
This process definitely needs more love. And I'm excited to see how translators can use html we send.


docs/TranslationWorkflow.md
```diff
@@ -0,0 +1,159 @@
...
+#### 4. Open your saved file in Chrome
+You should see a button at the top right that that reads *Translations Selected*
+
+#### 5. Highlight translations and download csv
+You can now click on items within the page to highlight them. This will indicate what sections need to be translated.
+As you click on items, you'll see a prompt which asks if you're sure about your selection. You may click an item to deselect it as well. 
+
+Once you've completed making selections, click the *Translations Selected* button. This will begin a download of two files. Your browser may ask for your permission to download multiple files. If so, click ok.
+
+Your files will be named something similar to **for-translations-1525922959252-index.html** and **for-translations-1525922959252.csv**
```
benweatherman: Is there any possibility to make these names a little more descriptive? Perhaps the slug of the URL? Using the date and time?

docs/TranslationWorkflow.md
```diff
@@ -0,0 +1,159 @@
...
+
+![Highlight content and Download the Page and CSV](./images/translation_workflow_5.gif)
```
benweatherman: Is it 1 CSV per translated page? I'm wondering if we should add a column for the filename the translations are attached to (but might be obvious from the name of the CSV). Disregard this if it doesn't make sense.

docs/TranslationWorkflow.md
```diff
@@ -0,0 +1,159 @@
...
+Once you've verified everything looks as expected, you can email or send this file as you would any other. Also, feel free to rename the file to something more user friendly.
+
+### Continue, if you require highlighting. 
+
+#### 3. Add code to page to allow highlighting
+Open your downloaded file in a text editor. I use SublimeText, but you can use any text editing application (TextEdit or Notepad are default applications for Macs and PCs respectively) that your computer may have.
+
+Paste the code below at the bottom of your file, and then save your file. 
+
+Note, I am not proud of this code. Do not write code like this for your own well-being and for preservation of your friendships.
```
mateoclarke: I am proud of this code.

docs/TranslationWorkflow.md
ifsimicoded: This is a great idea, but the export Chrome extension doesn't acknowledge the original filename, so I'd have to dig it up somewhere from the HTML. I'll investigate this and make updates (maybe use another chrome extension) and make a new PR if I find something simple to get the page slug.

PR #239, opened 2018-05-10
Add proptypes, i18n for header/footer storybook, misc

ifsimicoded:
- add proptypes (handles half of https://github.com/cityofaustin/techstack/issues/270)
- make group level proptype definitions for shared proptypes
- clean up data for primary navigation (handles some of https://github.com/cityofaustin/techstack/issues/244)
- decouple page layout from i18nController
- add I18n support for Header and Footer storybook components (https://github.com/cityofaustin/techstack/issues/273)
- remove coupled storybook components (Menu and MenuItems rely on the Header layout and parent controller functions)


src/components/PageSections/Menu/SubmenuItem.js
```diff
@@ -5,10 +5,20 @@ import I18nNavLink from 'components/I18n/I18nNavLink';
 
 const SubmenuItem = ({ className, topic, handleCloseAllMenus }) => (
   <li onClick={handleCloseAllMenus} className="coa-SubmenuItem" role="menuitem">
-    <I18nNavLink to={`/topics/${topic.slug}`} className={className}>
+    <I18nNavLink to={topic.url} className={className}>
```
ifsimicoded: same here, url building happens when the data is fetched in static.config.js

src/components/PageSections/Menu/_Submenu.scss
```diff
@@ -15,8 +15,7 @@
     & > li .coa-Submenu {
       left: 0;
     }
-    & > li:nth-last-of-type(2) .coa-Submenu,
-    & > li:last-of-type .coa-Submenu {
+    & > li:nth-of-type(n + 6) .coa-Submenu {
```
ifsimicoded: I did this to fix the right to left orientation for menuItem in storybook, but I ended up killing the menuItem storybook component. I left this, because it's simpler to understand.

src/components/Recollect/index.js
```diff
@@ -24,7 +25,7 @@ class Recollect extends Component {
   }
 
   render() {
-    const { intl } = this.props;
+    const { intl, options } = this.props;
```
ifsimicoded: just making the prop usage more obvious.

src/components/Tiles/TileGroup.js
```diff
@@ -41,10 +33,10 @@ const TileGroup = ({
 );
 
 TileGroup.propTypes = {
-  tiles: PropTypes.array.isRequired,
+  tiles: PropTypes.arrayOf(PropTypes.shape(Tile.propTypes)).isRequired,
   tag: PropTypes.string,
-  title: PropTypes.string,
-  titlePath: PropTypes.string,
+  text: PropTypes.string,
```
ifsimicoded: standardizing link formats so we can reuse the cleanLinks functions

.storybook/config.js
```diff
@@ -11,6 +12,13 @@ import localeMessages from 'js/i18n/loadLocaleData';
 
 import 'css/coa.css';
 
+//Fragment support hack (https://github.com/storybooks/addon-jsx/issues/34)
```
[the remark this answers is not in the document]
mateoclarke: what do you mean, then opted out?

src/components/I18n/I18nController.js
```diff
@@ -115,22 +110,16 @@ class I18nController extends Component {
...
-        <I18nDecorator>
-          <SkipToMain />
-          <LanguageSelectBar path={match.params.path || ''} />
-          <Header navigation={navigation[lang]} />
-          <main role="main" id="main">
-            <Routes />
-          </main>
-          <Footer threeoneone={threeoneone[lang]} />
-        </I18nDecorator>
+        <I18nDecorator>{children}</I18nDecorator>
```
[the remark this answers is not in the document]
mateoclarke: 👍

src/js/helpers/cleanData.js
```diff
@@ -49,15 +49,31 @@ export const cleanLinks = (links, pathPrefix) => {
...
     return {
       url: `${pathPrefix || ''}/${slug}`,
-      text: title,
+      text: title || text,
       ...rest,
     };
   });
 };
 
+export const cleanNavigation = navigation => {
```
mateoclarke: 💅

src/js/helpers/cleanData.js
```diff
@@ -49,15 +49,31 @@ export const cleanLinks = (links, pathPrefix) => {
...
+  const { allThemes } = navigation;
+
+  if (!allThemes || !allThemes.edges) return null;
+
+  let cleanedNavigation = cleanLinks(allThemes, '/themes');
+  cleanedNavigation.map(theme => {
```
mateoclarke: don't you need to set whatever the map returns to a new var so when you return it on line 74 its the updated array? mapping won't mutate `cleanedNavigation`. Or wait, the mutations you're doing are inside the object so they do persist? hmm, I'm probably wrong but maybe double check?

ifsimicoded: yep, the second half of what you're saying is what's going on -- doing mutations on the object directly and since its memory reference sticks, the change persists when the function exits.

.storybook/config.js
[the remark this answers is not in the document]
ifsimicoded: I was using Fragment in one of the components, I think the AppView in App.js, but then I just wrapped it in a div. So I'm not using fragments anywhere I don't think yet. but if we do, we'll need this bit of code.

src/components/Contact/ContactDetails.js
```diff
@@ -5,10 +5,17 @@ import { injectIntl } from 'react-intl';
...
-import Address from 'components/Contact/Address';
-import Hours from 'components/Contact/Hours';
+import Phone from './Phone';
+import Email from './Email';
+import Address from './Address';
+import Hours from './Hours';
+
+import {
+  addressPropTypes,
+  emailPropTypes,
```
luqmaan: I would call this `AddressPropTypes`, `EmailPropTypes`, etc.

src/components/Contact/proptypes.js
```diff
@@ -0,0 +1,25 @@
+import PropTypes from 'prop-types';
+
+export const addressPropTypes = PropTypes.shape({
+  name: PropTypes.string.isRequired,
+  street: PropTypes.string.isRequired,
+  city: PropTypes.string.isRequired,
+  state: PropTypes.string.isRequired,
+  zip: PropTypes.string.isRequired,
+}).isRequired;
```
luqmaan: 👍

PR #242, opened 2018-05-24
Feedback form on all page and more tech debt

ifsimicoded:
- add feedback form to all pages (Closes out https://github.com/cityofaustin/techstack/issues/348)
- finish up tech debt ticket about moving data transformations out of page level components (Closes out https://github.com/cityofaustin/techstack/issues/244)

### DIFF
--- a/src/components/PageSections/Footer/_Footer.scss
+++ b/src/components/PageSections/Footer/_Footer.scss
@@ -2,7 +2,7 @@
   background: $coa-color-purple-darkest;
   color: $coa-color-white;
   font-family: $coa-font-sans;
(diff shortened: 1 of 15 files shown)


mateoclarke:
just a heads up that you last commit broke your PR preview deploy @ifsimicoded

ifsimicoded:
Thanks!! It looks like there's something in our current setup that causes react-static's prod build to fail when using Fragments, so I adjusted to putting the feedback form inside the <footer> element.

ifsimicoded:
The Fragment issue might be related to having multiple versions of react as a dependency. https://github.com/facebook/react/issues/12097

Issue #245, opened 2018-06-02
Hack 4 Change Trash API


briaguya:
The goal is to replace the widget that we have on this page:
http://alpha.austin.gov/services/bulk-item-pickup

In order to implement this we need to:
- [ ] Create a component that takes in pickup schedule data and displays it in a responsive manner
- [ ] Create a component that allows a user to enter and select their address (with autosuggest/autocomplete)
- [ ] Figure out how to hook all of this up to the recollect API

briaguya:
## Recollect API

### Recycling schedule:
When typing in the address for autocomplete:
```https://recollect.net/api/areas/Austin/services/323/address-suggest?q=1702&locale=en-US&_=1527882407286```
When selecting an address, it appears to do a formatted address search to get an ID
```https://recollect.net/api/places?formatted_address=1702+Taylor+Gaines+St+Unit+B%2C+Austin&parcel_id=39514805&area_id=323&area_name=Austin&place_id=0&svc_id=323&locale=en-US&args=```
when hitting that endpoint, it redirects to:
```https://recollect.net/api/places/9155CF5C-65D4-11E8-891D-DF22E1657EA1/services/323?```
then we get a calendar search endpoint hit:
```https://recollect.net/api/areas/Austin/services/waste/pages/en-US/calendar_search:%7B%22tabbed%22:true%7D.json?widget_config=%7B%22area%22%3A%22Austin%22%2C%22container%22%3A%22%23rCw%22%2C%22page%22%3A%22tabbed_widget%22%2C%22name%22%3A%22wizard%22%7D&_=1527882494973```
^ this one doesn't seem to actually get any specific info
We also get a hit on this endpoint:
```https://recollect.net/api/places/9155CF5C-65D4-11E8-891D-DF22E1657EA1/services/323/events?nomerge=1&hide=reminder_only&after=2018-05-27&before=2018-07-09&locale=en-US&include_message=email&_=1527882518```

ifsimicoded:
@briaguya not sure the best place for this issue to live, but just FYI, https://github.com/cityofaustin/techstack/issues/ is where all the other janis/joplin issues live.

PR #251, opened 2018-06-03
Hack for Change Recycling Schedule

briaguya:
Closes https://github.com/cityofaustin/janis/issues/245


src/components/Trashy/index.js
```diff
@@ -0,0 +1,81 @@
...
+    this.setState({ address: address });
+    this.getAddressId(address);
+  };
+
+  render() {
+    const { intl, options } = this.props;
+    return (
+      <div className="coa-Trashy">
+        <TrashyAddress
+          suggestions={this.state.addressSuggestions}
```
ifsimicoded: Just curious why TrashyAddress doesn't manage fetching address suggestions and the returned list of addressSuggestions as well? It doesn't look like any other component needs to be aware of this data or fetching process but it's totally possible that I'm overlooking something

src/components/Trashy/index.js
```diff
@@ -0,0 +1,81 @@
...
+          setAddress={this.setAddress}
+          getSuggestions={this.getAddressSuggestions}
+        />
+        {this.state.address && (
+          <TrashySchedule
```
ifsimicoded: Also curious why TrashySchedule doesn't manage fetching pickupDates and nextBulkPickupDate on its own too. It doesn't seem like other components need awareness of this data or process. Looks like TrashySchedule only needs an addressId and on componentWillReceiveProps within TrashySchedule you'd trigger a fetch? Again, It's very likely that I'm overlooking something or that there's a better reason to structure React components this way.

src/components/Trashy/TrashySchedule.js
```diff
@@ -0,0 +1,105 @@
...
+    <I18nLink to="bulk-item-pickup">
+      {intl.formatMessage(i18n.bulkitemcollection)}
+    </I18nLink>
+  );
+
+  return (
+    <div className="coa-Trashy__schedule-container">
+      <div className="coa-Trashy__schedule-header">
+        <h3>
+          <FormattedMessage
```
ifsimicoded: Instead of using FormattedMessage, could we use intl.formatMessage where possible, so we can create these translation definitions in src/js/i18n/definitions.js so we have one central place for definitions that can easily be shared between components?

src/components/Trashy/TrashySchedule.js
```diff
@@ -0,0 +1,105 @@
...
+    {
+      var cases = {
+        textiles: intl.formatMessage(i18n.textiles),
+        recycling: intl.formatMessage(i18n.recycling),
+        yardtrimmings: intl.formatMessage(i18n.yardtrimmings),
+        trash: intl.formatMessage(i18n.trash),
+        compost: intl.formatMessage(i18n.compost),
+        brushcollection: intl.formatMessage(i18n.brushcollection),
+        bulkitemcollection: (
+          <I18nLink to="bulk-item-pickup">
```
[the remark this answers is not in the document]
briaguya: created an issue for this: https://github.com/cityofaustin/techstack/issues/443

src/components/Trashy/index.js
```diff
@@ -0,0 +1,81 @@
...
+  state = {
+    address: null,
+    addressSuggestions: [],
+    pickupDates: [],
+    nextBulkPickupDate: null,
+  };
+
+  getAddressSuggestions = addressText => {
+    axios
+      .post('https://recollect-wrapper.herokuapp.com/api/suggest', {
```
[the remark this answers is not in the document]
briaguya: made an issue: https://github.com/cityofaustin/techstack/issues/444

src/components/Trashy/index.js
briaguya: As we were building it out we didn't know which calls would give us what data, so we decided to try to abstract as much of the logic around fetching data from the api out of any of the display components, this also made it much easier to develop the schedule within storybook.

I'm not married to this implementation, and it's definitely a discussion I'm interested in having as we move towards a more robust implementation of the recollect api: https://github.com/cityofaustin/techstack/issues/444

PR #255, opened 2018-06-11
Process Page Badge Component

ifsimicoded:
This PR makes the badge and badge group components for the process page.
Closes https://github.com/cityofaustin/techstack/issues/399


src/components/Badges/Badge.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
```
mateoclarke: 💅

src/components/Badges/BadgeGroup.js
```diff
@@ -0,0 +1,18 @@
...
+import PropTypes from 'prop-types';
+
+import Badge from './Badge';
+
+import { badgeGroupPropTypes } from './proptypes';
+
+const BadgeGroup = ({ badges }) => (
+  <div className="coa-BadgeGroup">
+    {badges.map(({ ...rest }, index) => (
+      <Badge {...rest} />
```
mateoclarke: nice use of `...rest`

src/components/Badges/_Badge.scss
```diff
@@ -0,0 +1,83 @@
...
+
+.coa-Badge--isActive {
+  .coa-Badge__background {
+    path {
+      stroke: $coa-color-purple-dark;
+      stroke-width: $coa-badge-stroke + 2px;
+    }
+  }
+  .coa-Badge__title {
+    text-decoration-color: $coa-color-green-light; /* limited browser compat */
```
mateoclarke: what does this comment mean?

src/css/base/_variables.scss
```diff
@@ -94,6 +96,17 @@ $coa-tile-group-border: 4px;
 $coa-tile-group-margin: 1%;
 $coa-sectionheader-highlight-height: 8px;
 $coa-sectionheader-highlight-width: 34px;
+$coa-badge-height: 90px;
```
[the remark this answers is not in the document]
mateoclarke: Not sure if there is a right answer but my vote would be `rem` when possible.

Hot take: https://medium.com/@julienetienne/pixels-are-dead-faa87cd8c8b9
SO with the outdated answer being the most updated and a more recent answer being the second most upvoted: https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css

src/components/Badges/_Badge.scss
ifsimicoded: text decoration color doesn't really have broad browser compatibility so changing the underline color isn't simple. so we shouldn't expect the underline to be green in most cases. Open to rewording to make this comment more clear if you have suggestions.

PR #260, opened 2018-06-21
make page header work like section header for consistency

ifsimicoded:
I got bored waiting for Docker to build yesterday so I made some minor changes to make our headers components work consistently. I probably should've read a Medium post instead.

### DIFF
--- a/src/components/PageHeader/index.js
+++ b/src/components/PageHeader/index.js
@@ -2,9 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-const PageHeader = ({ title, description }) => (
+const PageHeader = ({ description, children }) => (
   <div className="coa-PageHeader">
-    <h1>{title}</h1>
+    <h1>{children}</h1>
     {description && (
       <p className="coa-PageHeader__description">{description}</p>
     )}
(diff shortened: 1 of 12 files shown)


src/components/PageHeader/index.js
```diff
@@ -2,17 +2,17 @@ import React from 'react';
...
+    <h1>{children}</h1>
     {description && (
       <p className="coa-PageHeader__description">{description}</p>
     )}
   </div>
 );
 
 PageHeader.propTypes = {
-  title: PropTypes.string.isRequired,
+  children: PropTypes.node.isRequired,
```
mateoclarke: is there any difference between `node` and `element`?

src/components/Pages/Departments.js
```diff
@@ -7,7 +7,7 @@ import TileGroup from 'components/Tiles/TileGroup';
 const Departments = ({ departments }) => (
   <div>
     <div className="wrapper wrapper--sm container-fluid">
-      <PageHeader title={'All Departments'} />
+      <PageHeader>All Departments</PageHeader>
```
mateoclarke: 💅

src/components/PageHeader/index.js
ifsimicoded: element I think expects one react element, node can be anything (string, number, arrays, fragments, an element, etc)

PR #262, opened 2018-07-11
CMS preview skeletion

ifsimicoded:
Added wrapper components for the /preview/ route and for the react-static production route.
Closes https://github.com/cityofaustin/techstack/issues/478


src/components/_Controllers/CMSLive.js
```diff
@@ -0,0 +1,16 @@
...
+import ReactGA from 'react-ga';
+import { logPageView } from 'js/helpers/googleAnalytics';
+
+const CMSLive = ({ intl }) => {
+  ReactGA.initialize(process.env.GOOGLE_ANALYTICS, { titleCase: false });
+  logPageView();
+  return <Routes />;
+};
+
+CMSLive.propTypes = {};
```
mateoclarke: should we put `intl` in here?

src/components/_Controllers/CMSPreview.js
```diff
@@ -0,0 +1,75 @@
...
+    const client = createGraphQLClientsByLang(intl.locale);
+    //TODO: one endpoint for revisions data requests
+    let req;
+    switch (page_type) {
+      case 'services':
+        req = client.request(allServicePagesQuery);
+        break;
+      case 'processes':
+        req = client.request(allProcessesQuery);
+        break;
```
mateoclarke: This switch statement is making me think it would be nice to have some documentation or checklist about what we need to change in Joplin whenever we add a new content type. Just a thought so we don't forget something in the future.

src/components/_Controllers/CMSPreview.js
```diff
@@ -0,0 +1,75 @@
...
+        <Route
+          path="/processes"
+          render={props => <Processes processes={cleanProcesses(data)} />}
+        />
+      </Switch>
+    );
+  }
+}
+
+CMSPreview.propTypes = {};
```
mateoclarke: should we add `intl` and 
```
match: {
  params: { revision_id, page_type },
}
```

src/js/helpers/fetchData.js
```diff
@@ -1,4 +1,13 @@
 import axios from 'axios';
+import { GraphQLClient } from 'graphql-request';
+
+export const createGraphQLClientsByLang = lang => {
```
mateoclarke: ~nice! I'm thinking whether this function would save us any code in the static.config.js file~

src/js/i18n/constants.js
```diff
@@ -57,6 +57,9 @@ export const SUPPORTED_LANGUAGES = [
 
 export const SUPPORTED_LANG_CODES = SUPPORTED_LANGUAGES.map(lang => lang.code);
 
+/* regex to find 2 character lang code (if present) */
+export const LANG_URL_REGEX = '(/)?:lang([a-z]{2})?/';
```
mateoclarke: 💅

src/js/helpers/fetchData.js
mateoclarke: derp, nvm. you did it already.

src/components/_Controllers/CMSLive.js
ifsimicoded: For HOC injection of a prop I didn't add it to propTypes since it's basically hard-coded as a prop, so it'll never not be there like other props that are fed in via data/react-static route json. And I hadn't done that for any other components with intl injected -- so just being consistent.

src/components/_Controllers/CMSPreview.js
ifsimicoded: I agree -- do you think the ReadMe is a good place for this?

src/components/_Controllers/CMSPreview.js
ifsimicoded: same as above, if the props are coming from an HOC or the route (in this case), they'll always be present b/c they're basically hardcoded. Open to change this tho.

src/components/_Controllers/CMSPreview.js
ifsimicoded: Actually, the more I think about it, we shouldn't preserve this switch statement long term as it couples our front end logic to the CMS -- we should rely on one request endpoint, that delivers the data formatted for the appropriate content-type. What do you think?

src/components/_Controllers/CMSLive.js
mateoclarke: word

src/components/_Controllers/CMSPreview.js
mateoclarke: ah yes, that does seem like a better long term solution.

briaguya:
Merged in via #263

PR #263, opened 2018-07-12
Use updated Joplin endpoint

briaguya:
Updated the preview component to use the new Joplin endpoint for page revisions


src/js/queries/getRevisionQuery.js
```diff
@@ -0,0 +1,64 @@
+const getServicePageRevisionQuery = `
+  query getPageRevision($id: ID) {
+    pageRevision(id: $id) {
+      id
+      asServicePage {
```
mateoclarke: nice!

src/components/_Controllers/CMSPreview.js
```diff
@@ -43,7 +48,19 @@ class CMSPreview extends Component {
         break;
     }
     req.then(data => {
-      this.setState({ data: data });
+      if (page_type === 'service') {
+        // Unitl we refactor cleanServices, reshape the object to work with it
```
mateoclarke: 👍

briaguya:
Works in staging now: https://janis-staging-pr-263.herokuapp.com/en/preview/service/UGFnZVJldmlzaW9uTm9kZToyODI=

package.json
```diff
@@ -29,7 +28,7 @@
     "react-load-script": "^0.0.6",
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
-    "react-static": "5.8.5",
+    "react-static": "^5.9.7",
```
ifsimicoded: just curious, why the update?

src/components/_Controllers/CMSPreview.js
```diff
@@ -35,6 +37,9 @@ class CMSPreview extends Component {
     //TODO: one endpoint for revisions data requests
     let req;
     switch (page_type) {
```
ifsimicoded: do you think we could get away without this switch somehow so we don't have to keep adding to it for new content-types?

package.json
briaguya: [because non-static routes were broken when running serve-build instead of serve-local](https://github.com/cityofaustin/janis/pull/263/commits/ad7cb36ec6fc1d2b324e0195554a9e1fb526e276)

src/components/_Controllers/CMSPreview.js
briaguya: I'm not sure, the way we generate the endpoints requires object types that will end up using different queries.
```
query getPageRevision($id: ID) {
  pageRevision(id: $id) {
    id
    asServicePage {
```

Unless we do some extensive refactoring on the backend our revisions will continue to be different objects types in graphQL and require different queries

PR #265, opened 2018-08-02
don’t show PageBanner without image prop

mateoclarke:
Closes https://github.com/cityofaustin/techstack/issues/574

@briaguya, when I showed you this at first, I was editing the PageBanner component to have the fallback of using a purplish background div when there is no image. But then I saw that our process pages had already solved this by just not rendering the PageBanner without an image prop so I applied that pattern to the Service page component too.

_service page with preview url rendering with no image in screenshot_
screen shot 2018-08-02 at 4 11 27 pm

### DIFF
--- a/src/components/Pages/Service.js
+++ b/src/components/Pages/Service.js
@@ -31,15 +31,17 @@ const Service = ({
   intl,
 }) => (
   <div>
-    <PageBanner
-      imagesPath={`${process.env.CMS_MEDIA}/images`}
-      imageFilename={path.basename(
-        image.filename,
-        path.extname(image.filename),
-      )}
-      imageExtension={path.extname(image.filename)}
-      imageTitle={image.title}
-    />
+    {image && (
+      <PageBanner
+        imagesPath={`${process.env.CMS_MEDIA}/images`}
+        imageFilename={path.basename(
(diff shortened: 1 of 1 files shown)


mateoclarke:
@ifsimicoded I thought the issue with revisions is that if an image wasn't yet chosen for a drafted page, the preview would break. That is what I took from your original comment: https://github.com/cityofaustin/techstack/issues/479#issuecomment-405643611

If images don't persist from revision to revision, that would be a separate bug. Is that something you've seen?

ifsimicoded:
I definitely wasn't clear in the issue. It seems like the image isn't persisted from revision to revision unless the image is selected via the Wagtail CMS interface. So if the image is selected, via our automated script that processes YAML files, the image won't be persisted with the revision when the next draft is published. You'll see in the CMS for the [Revisions of the Drop off household hazardous waste and other recyclables Page](http://joplin-staging.herokuapp.com/admin/pages/8/revisions/) -- that there is no image selected for past revisions.

mateoclarke:
@ifsimicoded Yes log it as a different issue. I'm not quite following but this seems like it is only an issue for seeded data in draft state. Do you have any lead on how we solve it?

PR #266, opened 2018-08-22
use steps the way joplin makes them now

briaguya:


src/js/queries/allServicePagesQuery.js
```diff
@@ -1,65 +1,16 @@
...
-                }
-              }
-            }
-          }
+          ...servicePageInfo
         }
       }
     }
   }
+  ${servicePageFragment}
```
ifsimicoded: 🙏

src/js/queries/servicePageFragment.js
```diff
@@ -0,0 +1,65 @@
...
+      id
+      slug
+      text
+      theme {
+        id
+        slug
+        text
+      }
+    }
+    serviceSteps {
```
ifsimicoded: From the looks of our inbound public health content, it seems like we need a title field and an unknown length of steps (which could include nested lists) here.

src/components/Pages/Service.js
```diff
@@ -50,11 +50,11 @@ const Service = ({
     <div className="wrapper wrapper--sm container-fluid">
       <PageHeader>{title}</PageHeader>
 
-      {steps &&
-        !!steps.length && (
+      {serviceSteps &&
+        !!serviceSteps.edges.length && (
```
ifsimicoded: Can we remove edges when we process this data in static.config.js -- I think our components should be graphql neutral. We have other clean functions that set this pattern.

src/js/queries/servicePageFragment.js
briaguya: I noticed the need for a title [when updating the yaml in joplin]((https://github.com/cityofaustin/joplin/pull/155/files#diff-9c5ea5128d161862f6213b2beb0cf3f2R12). Could you clarify what you mean by an unknown length of steps?

PR #267, opened 2018-08-22
update styles to better render public health content

ifsimicoded:


ifsimicoded:
closing because the steps module will be undergoing a deeper design change.

PR #272, opened 2018-10-15
Residential Parking Form (no babel-polyfill)

mateoclarke:
This PR continues the work the work of #271 but backtracks on making any changes to dependencies. It just prevents babel-polyfill from being loaded in the webpack entry.

PR #274, opened 2018-10-23
Fix deployment of ATD branch

briaguya:
After getting master updated to work with the newest joplin updates, I just merged master back into this branch. Then I found an issue with builds failing due to react-static using the constructor in the RPP component, so I moved that logic into componentDidMount.


mateoclarke:
I'll go ahead an merge this PR into ours (#272). @johnclary heads up that this will break that link I sent you so I'll pass you the new link once it deploys.

PR #283, opened 2019-02-20
Migrate colors

easherma:


src/components/PageSections/Footer/index.js
```diff
@@ -19,9 +19,9 @@ import { threeoneonePropTypes } from 'components/PageSections/ThreeOneOne/propty
 const Footer = ({ threeoneone, intl }) => (
   <footer>
     <div className="wrapper wrapper--sm container-fluid">
-      <FormFeedback />
+      {/*<FormFeedback />*/}
```
easherma: I'm just commenting these out for now because we may or may not be including them for OPO launch.

src/components/PageSections/Header/_Header.scss
```diff
@@ -1,6 +1,8 @@
 .coa-Header {
-  border-bottom: 1px solid $coa-color-purple-lighter;
+  border-bottom: 1px solid lighten( $coa-color-white, 10% );
```
easherma: not sure why we don't use these SASS features so I'm happy to refactor this to match other color variables

PR #286, opened 2019-02-21
Let's have a Department Page!

briaguya:


src/components/PageSections/Menu/_Menu.scss
```diff
@@ -8,39 +8,39 @@
...
-    bottom: auto;
-    left: auto;
-    margin: $coa-space-level5 auto 0;
-    overflow-y: visible;
-    position: relative;
-    top: auto;
-    width: 100%;
-    z-index: 2;
-  }
+  // @include media($coa-large-screen) {
```
easherma: im fine with commenting this out especially for now, but maybe drop a note about why it's commented out and what else is handling the behavior for now

src/components/Pages/Department.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { get } from 'lodash';
 import { withRouteData } from 'react-static';
 import path from 'path';
+import Parser from 'html-react-parser';
```
easherma: <3

easherma:
Haven't tested this, but LGTM! comments inline, only thing that sticks out is we may be able to re-use some of the department stuff for our new/upcoming menu component but we'll cross that bridge.

src/components/PageSections/Menu/_Menu.scss
briaguya: The menu button shows/hides when it's smaller than a large screen. This makes it so we stick with the menu button/collapsable menu instead of having the entire nav menu across the top when the screen is big enough

PR #324, opened 2019-03-07
1409 style page title desc

pibemillo:
- adds missing font-weights for IBM Plex Sans
- styles pageheader and adds section header
- rearranges info page, tweaks styles

### DIFF
--- a/src/components/HtmlFromAdmin/index.js
+++ b/src/components/HtmlFromAdmin/index.js
@@ -1,11 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import SectionHeader from 'components/SectionHeader';
-
 const HtmlFromAdmin = ({ title, content }) => (
   <div className="coa-HtmlFromAdmin">
-    {/* {title && <SectionHeader>{title}</SectionHeader>} */}
     <div
       className="coa-HtmlFromAdmin__content"
       dangerouslySetInnerHTML={{ __html: content }}
--- a/src/components/HtmlFromAdmin/index.js
+++ b/src/components/HtmlFromAdmin/index.js
@@ -5,7 +5,7 @@ import SectionHeader from 'components/SectionHeader';
 
 const HtmlFromAdmin = ({ title, content }) => (
   <div className="coa-HtmlFromAdmin">
-    {title && <SectionHeader>{title}</SectionHeader>}
+    {/* {title && <SectionHeader>{title}</SectionHeader>} */}
     <div
       className="coa-HtmlFromAdmin__content"
       dangerouslySetInnerHTML={{ __html: content }}
--- a/src/components/PageHeader/_PageHeader.scss
+++ b/src/components/PageHeader/_PageHeader.scss
@@ -1,8 +1,20 @@
 .coa-PageHeader {
-  margin-bottom: $coa-spacing-large;
-  padding-bottom: 1px; //prevent margin-stacking
+  // margin-bottom: $coa-spacing-large;
+  // padding-bottom: 1px; //prevent margin-stacking
 
   h1 {
-    text-align: left;
+    color: $coa-color-dark-blue;
+    font-size: $coa-font-size-xlarge;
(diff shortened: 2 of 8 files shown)


src/components/HtmlFromAdmin/index.js
```diff
@@ -5,7 +5,7 @@ import SectionHeader from 'components/SectionHeader';
 
 const HtmlFromAdmin = ({ title, content }) => (
   <div className="coa-HtmlFromAdmin">
-    {title && <SectionHeader>{title}</SectionHeader>}
+    {/* {title && <SectionHeader>{title}</SectionHeader>} */}
```
easherma: I would like an explanation for why code is commented out, and/or just remove it if we don't have plans to put this back into the code.

src/components/PageHeader/_PageHeader.scss
```diff
@@ -1,8 +1,20 @@
 .coa-PageHeader {
-  margin-bottom: $coa-spacing-large;
-  padding-bottom: 1px; //prevent margin-stacking
+  // margin-bottom: $coa-spacing-large;
```
easherma: same here, does this style get reflected elsewhere, or is it no longer relevant?

src/css/base/_variables.scss
```diff
@@ -1,5 +1,5 @@
 //typography
-@import url('https://fonts.googleapis.com/css?family=IBM+Plex+Sans:400,400i|Open+Sans:400,600,700,800');
+@import url('https://fonts.googleapis.com/css?family=IBM+Plex+Sans:400,400i,600,700|Open+Sans:400,600,700,800');
```
easherma: Nice catch!

src/css/base/_variables.scss
```diff
@@ -9,7 +9,7 @@ $coa-font-serif: $coa-font-sans;
 $coa-font-regular: 400;
 $coa-font-semibold: 600;
 $coa-font-bold: 700;
-$coa-font-extrabold: 800;
+$coa-font-extrabold: 700;
```
easherma: we should talk about this...I don't like having the same font-weight being in two variables...can we audit where these variables are used? and/or continue to have a font weight of 800 available to use (maybe this isn't possible with IBM Plex Sans, though)

src/css/base/_variables.scss
```diff
@@ -39,13 +39,13 @@ $coa-xlarge-screen: 1250px;
 
 //padding / margin
 $coa-space-level1: 0.5rem;
-$coa-space-level2: 1.0rem;
+$coa-space-level2: 1rem;
```
easherma: not considering this blocking, but there are some style tweaks on this page, not sure if we have a consistent style guide for unit numbers, color hashes, etc

src/css/base/_variables.scss
pibemillo: IBM Plex Sans maxes out at boldness/font-weight 700. I can audit and replace instances of $coa-font-extrabold with $coa-font-bold.

src/components/HtmlFromAdmin/index.js
pibemillo: Removed

src/components/PageHeader/_PageHeader.scss
pibemillo: It's no longer relevant. There are a some stylistic modifiers/differences for the PageHeader component depending on the content-type (service, information, topic) that will have to be accounted for soon.

src/css/base/_variables.scss
pibemillo: I'll have to connect with Chase to see if that's ready/available because it'll definitely have to get updated here

src/components/PageHeader/_PageHeader.scss
easherma: sounds good; maybe we should stick a brief note explaining that?

src/css/base/_variables.scss
easherma: thumb up emoji

PR #325, opened 2019-03-07
Issue 1331: make time dumber

Niicck:
Resolves cityofaustin/techstack#1331

Also added a "yarn start-local" script which modifies ./scripts/serve-local.sh to point to Production. I'm running off ctxfloods naming convention, feel free to suggest a better name.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,9 @@
     "storybook-addon-intl": "^2.3.0"
   },
   "scripts": {
-    "start-local": "./scripts/serve-local.sh -P",
+    "start-local": "./scripts/serve-local.sh",
+    "start-staging": "./scripts/serve-local.sh -S",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -51,6 +51,7 @@
     "storybook-addon-intl": "^2.3.0"
   },
   "scripts": {
+    "start-local": "./scripts/serve-local.sh -P",
```
sergiogcx: Thank you for adding this code, I think it makes more sense from a new user standpoint to use yarn/npm for this.

This script is going to default to pulling data from prod (-P), not sure if this the best way forward. We are creating two ways of using the app, one is through ./scripts/serve-local.sh and another through npm which defaults to prod.

Would there be a way to make prod optional? 

I am totally fine either way 👍

Niicck: Added scripts to point to staging, local, and prod

PR #326, opened 2019-03-11
mapping headings down to h4

easherma:
working on https://github.com/cityofaustin/techstack/issues/1423 brought up some larger aspects of how content is styled once joplin spits it over

i feel like the changes are sorta self-evident, but I'm also stubbing this out so we can discuss it during sync


easherma:
Great to hear that! Part of why I left things the way I did is because I
was thinking/hoping that would be true.

PR #329, opened 2019-03-12
1408: load director headshots from database

Niicck:
Resolves cityofaustin/techstack#1408
Additional documentation available on Joplin PR: https://github.com/cityofaustin/joplin/pull/200

### DIFF
--- a/src/components/DirectorHeadshot/index.js
+++ b/src/components/DirectorHeadshot/index.js
@@ -1,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import path from 'path';
+
+import ResponsiveImage from 'components/ResponsiveImage';
+
+const DirectorHeadshot = ({ photo }) => {
+  const filename = photo.filename;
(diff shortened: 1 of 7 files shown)


src/components/DirectorHeadshot/index.js
```diff
@@ -0,0 +1,35 @@
...
+  );
+  const filepath = `${process.env.CMS_MEDIA}/images/${photoBasename}`;
+
+  return (
+    <div className="coa-DepartmentPage__directorcard-headshot">
+      <ResponsiveImage
+        className="coa-DirectorHeadshotImage"
+        filename={filepath}
+        defaultWidth="width-640"
+        widths={[]}
```
briaguya: What happens if this is unset instead of an empty array?

Niicck: That was intentional. I think the page broke or the image didn't show up if widths (and by extension srcSet) was undefined.

PR #331, opened 2019-03-14
use relative URLs; adds siteRoot to config (not sure if needed atm)

easherma:


static.config.js
```diff
@@ -185,6 +185,8 @@ const makeDepartmentPages = async client => {
 };
 
 export default {
+  siteRoot: 'https://alpha.austin.gov',
```
pibemillo: This is good to have, although I'm not sure how to call it elsewhere. Is it accessible using withRouteData from react-static?

easherma:
I’m still not clear on that either, to be honest

easherma:
I have not bucketed it.

PR #347, opened 2019-04-01
Janis S3 Pipeline

sergiogcx:


static.config.js
```diff
@@ -182,6 +182,7 @@ const makeDepartmentPages = async client => {
 
 export default {
   // siteRoot: 'https://alpha.austin.gov',
+  //basePath // Do not alter this line if you want a working PR
```
briaguya: Any chance we can pass a variable in here?

sergiogcx: I've repeatedly have had issues with webpack and environment variables, I will give it another try with environment variables.

sergiogcx: Fixed with env var, merging to master and see how it behaves there. This should not affect production.

PR #349, opened 2019-04-03
Fix Links

briaguya:
Implement quick fix for broken link internationalization, by replacing i18n links with react-static links. I found that we keep the language selection while the address bar loses it, but the links work this way.?

Not a perfect solution, but should do the trick in the short term. More info here: https://github.com/cityofaustin/techstack/issues/1649


sergiogcx:
It looks like the PR is working for me:
https://janis-pr.austintexas.io/349/

PR #362, opened 2019-04-16
Build topic pages

briaguya:
Topic pages should now look something like what we have designed in: https://xd.adobe.com/spec/3693e3d0-1e3d-4c07-6555-b09fccd75568-40a2/screen/1de14141-92a4-4e3e-b9bb-27961a97c72b/Leaf-node-MVP-Mobile-/

Check it! https://janis-pr.austintexas.io/362/health-safety/healthcare-prevention/

### DIFF
--- a/src/components/Pages/Topic.js
+++ b/src/components/Pages/Topic.js
@@ -25,24 +25,15 @@ const Topic = ({
     <Head>
       <title>{title}</title>
     </Head>
-    <PageBreadcrumbs parent={{ ...theme, subpath: 'themes' }} title={title} />
-    <div className="wrapper wrapper--sm container-fluid">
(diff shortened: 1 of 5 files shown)


src/components/Pages/Topic.js
```diff
@@ -11,23 +11,27 @@ import SectionHeader from 'components/SectionHeader';
 import TileGroup from 'components/Tiles/TileGroup';
 
 const Topic = ({
-  topic: { theme, text: title, description, callToAction, tiles },
+  topic: {
+    theme,
+    text: title,
+    description,
+    callToAction,
```
pibemillo: Think I saw you remove callToAction from Joplin, no harm having it in there for now and removing later

PR #381, opened 2019-05-02
1803 dept page related content

pibemillo:
https://github.com/cityofaustin/techstack/issues/1803

https://xd.adobe.com/spec/3693e3d0-1e3d-4c07-6555-b09fccd75568-40a2/screen/9055635c-de21-497d-b8fb-631c3e39c802/Department-MVP-Desktop/

### DIFF
--- a/src/components/Tiles/TopServices.js
+++ b/src/components/Tiles/TopServices.js
@@ -6,7 +6,7 @@ import ArrowRight from 'components/SVGs/ArrowRight';
 import Tile from './Tile';
 
 const TopServices = ({ tiles, title, url, locale, extraClasses }) => (
-  <div className={'coa-TopServices ' + (extraClasses ? extraClasses : '')}>
+  <div className={`coa-TopServices ${extraClasses ? extraClasses : ''}`}>
     <div className="wrapper container-fluid">
       {title &&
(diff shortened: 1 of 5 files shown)


src/components/Tiles/TopServices.js
```diff
@@ -6,16 +6,17 @@ import ArrowRight from 'components/SVGs/ArrowRight';
 import Tile from './Tile';
 
 const TopServices = ({ tiles, title, url, locale, extraClasses }) => (
-  <div className={`coa-TopServices ${extraClasses}`}>
+  <div className={'coa-TopServices ' + (extraClasses ? extraClasses : '')}>
```
briaguya: wouldn't this work as:
`coa-TopServices ${!!extraClasses && extraClasses}`
or just
`coa-TopServices ${extraClasses ? extraClasses : ''}`

Is there something about [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) that's causing issues?

pibemillo: Updated the template literal to your second snippet. The others were returning false or undefined, so it was just a cleanup. Thanks for spotting!

PR #384, opened 2019-05-02
adds z-index to is-open

pibemillo:
https://github.com/cityofaustin/techstack/issues/1805


pibemillo:
When HowYouKnow menu is open, it is no longer cut-off or hidden behind the nav. Quick z-index fix

PR #394, opened 2019-05-08
hacky and hardcoded but gets us somewhere

briaguya:


easherma:
@briaguya I added another hacky option, but I think this is worth merging for convenience. We can skip docker and build either from joplin prod or from a local running instance of joplin.

easherma:
related: https://github.com/cityofaustin/techstack/issues/1869

easherma:
builds on netlify

PR #398, opened 2019-05-14
temp container, add actual grey color from design specs, relative + z-index

easherma:
* the gray lightest dosen't _quite_ match the grey @chasechenevert had in his spec. I added it to variables with a temp variable name for discussion. 
* I just made a temp container to get the background color in place, as it looks like we may be altering the home page to add in the footer(s) so this may be temporary

### DIFF
--- a/src/components/HeroHome/_HeroHome.scss
+++ b/src/components/HeroHome/_HeroHome.scss
@@ -1,7 +1,7 @@
 .coa-HeroHome__container {
   img {
     position: absolute;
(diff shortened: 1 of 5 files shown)


easherma:
This broken on mobile atm.

easherma:
This is improved; I made some inline styles where needed because a) the home page is probably exceptional and b) to kinda force the issue of discussing where those sorts of style changes might live

PR #399, opened 2019-05-14
fixes some incorrect JSX syntax i noticed

easherma:


easherma:
@sergiogcx I noticed some minor jsx things with https://github.com/cityofaustin/janis/pull/397 that are fixed here, but please review in case that was intentional for some reason I'm missing

PR #402, opened 2019-05-14
Updated tracking environment variable

sergiogcx:
Updates the environment variable being defined in staging for the google tracking code. Basically we need to track production only, not the other environments.


sergiogcx:
It looks like it is working, it does not affect the rest of the site, only staging and PRs.

PR #407, opened 2019-05-17
makes things work when we have everything but OPO unpublished

briaguya:


briaguya:
Here's a build I made against a local instance of Joplin with all the pages unpublished: http://janis-test-out-unpublished-stuff.s3-website-us-east-1.amazonaws.com/en/police-oversight

PR #408, opened 2019-05-21
when plain phone number is on the backend, make it a conforming object

easherma:
this seemed like the most non-destructive/incremental way to patch this
Fixes https://github.com/cityofaustin/techstack/issues/1950

### DIFF
--- a/src/js/helpers/cleanData.js
+++ b/src/js/helpers/cleanData.js
@@ -26,7 +26,6 @@ export const cleanContacts = contacts => {
         cleaned.phone = JSON.parse(cleaned.phone);
       } catch (error) {
         cleaned.phone = JSON.stringify({ default: cleaned.phone });
-        console.log('fix phone error', cleaned.phone);
       }
     }
     if (cleaned.hours && cleaned.hours.edges) {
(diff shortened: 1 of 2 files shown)


src/js/helpers/cleanData.js
```diff
@@ -20,8 +20,13 @@ export const cleanContacts = contacts => {
     // Yes, it's `contact.contact` because of the way the API returns data
     let cleaned = Object.assign({}, contact.contact);
 
-    if (cleaned.phone) {
-      cleaned.phone = JSON.parse(cleaned.phone);
+    // quick fix re: phone numbers are current inconsiently entered on the backend
+    try {
+      if (cleaned.phone) {
```
briaguya: maybe it makes sense to put this line outside the try block. I don't remember how our contact component handles null/empty for phone vs how it'll handle json with `default: null` or `default: ''`

easherma: tweaked in: https://github.com/cityofaustin/janis/pull/408/commits/dc59c4a37a7c5b4d9dee3acf9945cc097dfe0046

PR #409, opened 2019-05-22
1903 nav footer

briaguya:
We have the data getting in there, should be mostly just some css and react cleanup at this point.


src/components/PageSections/Footer/_Footer.scss
```diff
@@ -128,29 +128,71 @@
   }
 }
 
-.coa-FooterSiteMap__title {
-  font-size: $coa-font-size-small;
-  line-height: $coa-line-height-small;
-  font-weight: $coa-font-medium;
-  margin-bottom: $coa-spacing-small;
+.coa-FooterSiteMap {
+  display: none;
```
easherma: Whats the point of this? Controlling the display at different breakpoints?

src/components/PageSections/Menu/ThemesTopicsMenu.js
```diff
@@ -31,7 +31,11 @@ const TopicsLinks = props =>
 
 const ThemesTopicsMenu = props => {
   return (
-    <nav className="coa-ThemesTopicsMenu">
+    <nav
+      className={`coa-ThemesTopicsMenu ${
```
easherma: A note on what extraClasses is or intended use would be cool, though I can see from the footerSiteMap what the intended use is.

PR #416, opened 2019-05-23
Fix menu close bug

briaguya:
Now clicking the 'X' works again


src/components/PageSections/Header/index.js
```diff
@@ -54,15 +54,21 @@ class Header extends Component {
     // Full site
     if (this.wrapperRef && !this.wrapperRef.contains(event.target)) {
       // If we're clicking on the mobile close button, we'll handle this in toggleFullSiteMenu instead
-      if (event.target.className !== 'coa-Header__menuIcon') {
+      if (
+        event.target.className !== 'coa-Header__menuIcon' &&
```
briaguya: This line was getting 'material-icons' when we clicked the X, clicking "menu" would still get us 'coa-Header__menuIcon'

PR #417, opened 2019-05-23
Updates to top menu style

sergiogcx:


src/components/PageSections/Header/_Header.scss
```diff
@@ -5,6 +5,12 @@
   color: $coa-color-white;
 }
 
+.coa-Header--container {
+  max-width: $coa-top-menu-max-width;
+  margin-left: auto !important;
```
easherma: can we be more specific to avoid using the !important tag? or at least leave a note to refactor later

briaguya:
resolves https://github.com/cityofaustin/techstack/issues/2013

src/components/PageSections/Header/_Header.scss
briaguya: just curious, what was forcing a different margin on this that makes us need it to be !important?

src/components/PageSections/Header/index.js
```diff
@@ -144,71 +144,74 @@ class Header extends Component {
...
-        />
-        <div className="coa-Header__mobile-languages">
-          <LanguageSelectBar path={path} />
-        </div>
-        <div className="coa-Header__container">
-          <div className="coa-Header__controls">
-            <div className="coa-Header__left-controls">
-              <div className="coa-Header__desktop-languages">
-                <LanguageSelectBar path={path} />
+        <div className="coa-Header--container">
```
easherma: 'container' seems like a confusing modifier name for me. at first glance it looks like the purpose of the container is to set a max width and center the content.

src/components/PageSections/HowYouKnowMenu/_HowYouKnowMenu.scss
```diff
@@ -3,7 +3,8 @@
   display: none;
   position: fixed;
   top: 0;
-  width: 100vw;
+  width: 100%;
+  max-width: $coa-top-menu-max-width;
```
easherma: so here were setting max width with the later defined variable as part of this component. which is cool. 
for the header above we add a container modifier that does this + centers with the auto margins?

PR #432, opened 2019-05-31
2214 dept page display issues

briaguya:
fixes https://github.com/cityofaustin/techstack/issues/2214

# Before:
Screen Shot 2019-05-31 at 2 46 48 PM
Screen Shot 2019-05-31 at 2 47 06 PM

# After:
Screen Shot 2019-05-31 at 2 45 58 PM
Screen Shot 2019-05-31 at 2 45 44 PM

### DIFF
--- a/src/components/PageBanner/_PageBanner.scss
+++ b/src/components/PageBanner/_PageBanner.scss
@@ -56,3 +56,16 @@
     margin-top: $coa-space-level6;
   }
 }
+
+.coa-PageBanner__container {
+  img {
+    position: absolute;
+    z-index: -1;
+    top: auto;
+    height: 30rem;
+    width: 100%;
+    object-fit: cover;
+    // This is for https://www.npmjs.com/package/object-fit-images
+    font-family: 'object-fit: cover;';
+  }
+}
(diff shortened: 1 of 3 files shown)


src/components/PageBanner/_PageBanner.scss
```diff
@@ -56,3 +56,16 @@
     margin-top: $coa-space-level6;
   }
 }
+
+.coa-PageBanner__container {
+  img {
+    position: absolute;
+    z-index: -1;
+    top: auto;
+    height: 30rem;
```
easherma: any way to use one of the height variables here? https://github.com/cityofaustin/janis/blob/7d109b4e6b6c70aab784585f37d1a9539edad7cf/src/css/base/_variables.scss#L138
Or if we expect it to vary from the hero on the home page, we could add a height variable in order to be consistent?

src/components/PageBanner/index.js
```diff
@@ -4,7 +4,7 @@ import ResponsiveImage from 'components/ResponsiveImage';
 import { FULL_WIDTH_RESPONSIVE_IMAGE_SIZES } from 'js/helpers/constants';
 
 const PageBanner = ({ imagesPath, imageFilename, imageExtension, imageTitle, headerText, extraClasses }) => (
-  <div className="coa-HeroHome__container">
+  <div className="coa-PageBanner__container">
```
easherma: Would it make more sense to call it HeroPage or something similar to have consistency between component names? This kinda ties into a larger issue around the naming/structure of existing Janis components though...

briaguya: I was following the naming of the component itself. If anything, I think we should get rid of HeroHome and make PageBanner more flexible so we can use it in HeroHome.

briaguya: Made an issue to address more of this (https://github.com/cityofaustin/techstack/issues/2270) merging this PR now

PR #456, opened 2019-07-11
Multiple departments per service page

briaguya:


briaguya:
@easherma this is very WIP but so far I have:
http://localhost:3000/police-oversight/service-page-test-multi-departments - works
http://localhost:3000/test-dept/service-page-test-multi-departments - works
http://localhost:3000/learn-more-about-austin-resource-recovery/service-page-test-multi-departments - doesn't work because we don't publish children of unpublished departments

briaguya:
http://localhost:3000/en/health-safety/health-records-certificates-2/birth-death-certificates/service-page-test-multi-departments also works, but doesn't show any offered by right now, working on a quick fix for that

briaguya:
Ok, updated it so we have the contextualnav showing multiple departments now:

Still have a few things that need to happen:
* Right now we're adding the pages to the "related information" link list on dept pages, we'll probably need to figure out how we want to handle where the links actually go (and how to meld them with the exisiting "drop a url in" top services structure)
* We don't have any decent spacing between links to different depts, I was going for the "hey this works!" implementation not a final one

PR #457, opened 2019-07-18
2479 buttons in steps

easherma:
hey this works alright!
Didn't require any backend changes, just some judicious hacking and reading of the html-parser docs
Fixes https://github.com/cityofaustin/techstack/issues/2479 , in terms of stepsWithOptions at least

### DIFF
--- a/src/components/Steps/StepWithOptions.js
+++ b/src/components/Steps/StepWithOptions.js
@@ -35,7 +35,8 @@ const StepOption = ({ option_name, option_description }) => (
           {Parser(option_description, {
             replace: domNode => {
               if (domNode.data === 'Start') {
-                return <span class="usa-button-primary">Start</span>;
+                return <nav class="usa-button-primary">Start</nav>;
+                console.dir(domNode);
               }
             },
           })}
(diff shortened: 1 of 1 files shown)


src/components/Steps/StepWithOptions.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Parser from 'html-react-parser';
+import Parser, { domToReact } from 'html-react-parser';
```
briaguya: Is there a reason we need to import `domToReact` here? Doesn't seem to be something we're using

easherma: Ah no I was trying previously but forgot to remove

easherma:
Next steps:
Figure out what publishing the service page might look like
Probably remove/modify some routing somewhere to forms.austin.gov on the OPO department page
Test with more than one service page, see if the fix is needed for other kinds of steps

briaguya:
Screen Shot 2019-07-17 at 11 30 35 PM
Screen Shot 2019-07-17 at 11 30 42 PM

I'm pretty sure we'll need to be able to support both buttons and standard link style links in there. @chasechenevert any design input?

toribr:
@briaguya Yes, you are correct. We don't want all the links converted to buttons. In fact, the only button that should be converted is the "Start" button in the first option.

easherma:
Lets close this, we went with a different route

PR #459, opened 2019-07-23
Make LoADding not messy

briaguya:


toribr:
I dislike this implementation altogether. The icons are ugly, and having it embedded in the page seems weird to me. Can we not just use a css loader instead of something inline?

PR #493, opened 2019-08-28
Department top services

briaguya:


src/js/queries/departmentPageFragment.js
```diff
@@ -26,7 +26,119 @@ const departmentPageFragment = `
...
+            relatedDepartments {
+              edges {
+                node {
+                  relatedDepartment {
+                    slug
+                  }
+                }
+              }
+            }
+            topics {
```
briaguya: this can be cleaned by https://github.com/cityofaustin/techstack/issues/2851

src/components/Pages/Department.js
```diff
@@ -15,9 +15,6 @@ import PageBreadcrumbs from 'components/PageBreadcrumbs';
 import PageHeader from 'components/PageHeader';
 import WorkInProgress from 'components/WorkInProgress';
 
-// TODO: this jsonFileData is temporary. Add it to Wagtail API
```
easherma: What was this?

package.json
```diff
@@ -59,7 +59,7 @@
   },
   "scripts": {
     "start-local": "./scripts/serve-local.sh",
-    "start-joplin-cloud": "NODE_PATH=./src CMS_API='http://joplin.herokuapp.com/api/graphql' CMS_MEDIA='https://joplin-austin-gov.s3.amazonaws.com/media' npm-run-all -p watch-css start-js",
```
easherma: Lets change this back to default joplin prod before merge/maybe we should make it easier to run this command off of different branches

static.config.js
```diff
@@ -295,14 +297,22 @@ const makeDepartmentPages = async client => {
...
+
     infoPage.type = 'info';
 
     let matchingDepartmentIndex = departments.findIndex(
       d => d.id === infoPage.department.id,
     );
 
     if (departments[matchingDepartmentIndex]) {
-      departments[matchingDepartmentIndex].relatedLinks.push(infoPage);
+      // Only add the related link if it's not a top service
```
easherma: this conditional block (308-315) seems almost identical to the one (341-346), can we make it more clear what is going on here?

src/js/helpers/cleanData.js
```diff
@@ -283,7 +283,73 @@ export const cleanDepartmentDirectors = directors => {
   });
 };
 
-export const cleanDepartments = allDepartments => {
+export const cleanDepartmentTopServiceLink = (topService, langCode) => {
```
easherma: I haven't tested this yet and it looks solid, but it feels like a lot of work for the frontend to have to do just for making URLs.

static.config.js
briaguya: This one is for information pages, the other is for service pages, they end up as different objects here

src/js/helpers/cleanData.js
briaguya: it is less than ideal, but it isn't far removed from our existing logic around making links work. There's a bit extra in here to make sure we retain the page IDs so we can avoid having any top services show up as related information.

Since related information links are populated by looking through every page and finding parent departments during the frontend build process, as opposed to being managed in Joplin, this seemed like the best place for the logic.

This type of thing is a great example of why https://github.com/cityofaustin/techstack/issues/2792 exists

PR #504, opened 2019-09-11
2972 - support hosted documents

briaguya:
resolves the janis part of https://github.com/cityofaustin/techstack/issues/2972
includes support for multiple sources for documents (comments in code explaining why), example here: https://deploy-preview-504--janis.netlify.com/en/health-safety/healthcare-prevention/disease-prevention/test-docs


src/js/helpers/cleanData.js
```diff
@@ -460,23 +459,41 @@ export const clean311 = threeoneone => {
...
+  // Single source mode, example use case:
+  // If we're on PROD, we should only get prod documents
+  if (process.env.CMS_DOCS !== 'multiple') {
+    return `${process.env.CMS_DOCS}/${filename}`;
+  }
+
+  // Multi source mode, let's do some url checking and get something
+  // that works. example use case:
+  // If we're on STAGING, we want docs imported from PROD to work,
+  // as well as any new docs we added when testing on staging
```
Niicck: Ok. So this allows us to not need to datadump documents as well, correct? But still look at prod documents on staging, but not look at staging documents on prod.

Alternatively, the document saved in the db could include its root url/CMS_DOCS location.

src/js/helpers/cleanData.js
```diff
@@ -485,16 +502,28 @@ export const cleanOfficialDocumentPages = async allOfficialDocumentPages => {
...
+      // use that info to update the information syncronously
+      if (doc.node.document) {
+        doc.node.link = await getWorkingDocumentLink(
+          doc.node.document.filename,
+        );
+
+        // Maybe there's a better way to handle this but meh for now
+        // If it's a pdf, add the size
+        if (doc.node.document.filename.slice(-3) === 'pdf') {
+          doc.node.pdfSize = filesize(doc.node.document.fileSize).replace(
```
Niicck: oh dang filesize isn't async lol

src/js/helpers/cleanData.js
Niicck: Another thought if we keep a CMS_DOCS == 'multiple' flag: could this just be a trigger based on NODE_ENV?

At some point it'd be great to tie environment variables to NODE_ENV and have an organized way to do that. But today is probably not that day.

PR #506, opened 2019-09-11
2938 - open guide page links in new tabs

briaguya:
resolves https://github.com/cityofaustin/techstack/issues/2938


src/components/Pages/Guide/GuideSectionCollection.js
```diff
@@ -64,14 +58,15 @@ function GuideSection({
             departments,
             slug,
           })}
+          target="_blank"
```
briaguya: this is the only thing that actually changed, everything else is just jsprettier formatting

briaguya:
went ahead and merged this into master because it's a one line fix

PR #515, opened 2019-09-13
2904 centered TOC btn andriod

robbobfrh84:
fixes: https://github.com/cityofaustin/techstack/issues/2904
deployed: https://deploy-preview-515--janis.netlify.com/

What I did: 
- It looks like android devices want to add their own padding to "row" classes. 
- setting the right margin to 0 looks to clean this up and won't effect IOs or other browsers.

### DIFF
--- a/src/components/Pages/Guide/_Guide.scss
+++ b/src/components/Pages/Guide/_Guide.scss
@@ -259,6 +259,10 @@ $max-guide-tablet-width: $coa-medium-screen;
   width: 100%;
 }
 
+.coa-GuidePage__content-container > .wrapper > .row {
+  margin-right: 0 !important;
+}
+
 .coa-GuideMenu__mobile-button {
   position: relative;
   display: flex;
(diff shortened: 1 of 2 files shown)


src/components/Pages/Guide/_Guide.scss
```diff
@@ -259,6 +259,10 @@ $max-guide-tablet-width: $coa-medium-screen;
   width: 100%;
 }
 
+.coa-GuidePage__content-container > .wrapper > .row {
```
briaguya: it is strange that this was getting set to -1 before, probably would be nice to figure out where our bootstrappy things are causing wonkyness/one-offy css.

robbobfrh84: I found where someone, at sometime, added this note. I've added a reference to that `TODO` within the css as well. 
https://github.com/cityofaustin/janis/blob/0cba7da41b4d6b637becb7b05f73255104a37528/src/components/Pages/Guide/_Guide.scss#L72

robbobfrh84:
I've added margin-right on the `.coa-GuidePage__main-content` as well to fix other right margin issues that were less specific to android. Turns out it wasn't just an android issue as @briaguya pointed out. 
-
Also, this didn't need `!important` after all. I'm realizing I need to "double-reload" to trigger the most recent changes. I'll look into improving that in my dev env.

PR #524, opened 2019-09-20
2892 Guide content is broken on IE 11

robbobfrh84:
This fixes #2892 Guide errors on IE 11
- Issue: https://app.zenhub.com/workspaces/techstack-5a78b88e1ce69f3510b678ef/issues/cityofaustin/techstack/2892
- [x] - IE 11 Guide positioning error
  - IE is finicky with `display: flex`. `block` for just IE 11 will patch this error. 
- [x] - IE 11 Guide tracking tabs error
  - IE draws `undefined` for window.scrollY. Use `window.pageYOffset` instead

Additionally.. I build a script for IE testing!
- This will add the script command with yarn: `yarn build-local`, which will build, serve and alert developer that it's done, because it takes so long....
  - @Niicck feel free to wipe/modify this in your PR build adventures

- Dependency required: `http-server`
  - You can install simply with npm by running `http-server -g http-server`

Reference: 
- https://developer.mozilla.org/it/docs/Web/API/Window/scrollX

### DIFF
--- a/src/components/Pages/Guide/index.js
+++ b/src/components/Pages/Guide/index.js
@@ -59,6 +59,7 @@ function Guide(props) {
     We want the GuideSection right before the first GuideSection that is past the window's position.
   **/
   function handleScroll() {
+    // printSections(sectionLocations, window.pageYOffset)
     let i = 0;
     while (i < sectionLocations.length) {
       if (window.pageYOffset < sectionLocations[i].offsetTop - 1) {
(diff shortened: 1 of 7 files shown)


src/components/Pages/Guide/index.js
```diff
@@ -58,17 +59,15 @@ function Guide(props) {
     We want the GuideSection right before the first GuideSection that is past the window's position.
   **/
   function handleScroll() {
-    const thing = sectionLocations[0].offsetTop;
```
robbobfrh84: @nick looks like this was a ruin of your testing, so I "put it out to pasture". 🌅

nick: you mean @Niicck

robbobfrh84: Sorry @Nick! I'll look out for that in the future.
- But, you have a great Friday evening, ok!

PR #531, opened 2019-10-01
3037 fix tile group title url localization 

briaguya:
once we move away from old react-static it'll be great to clean up links
fixes https://github.com/cityofaustin/techstack/issues/3037

### DIFF
--- a/src/components/Tiles/TileGroup.js
+++ b/src/components/Tiles/TileGroup.js
@@ -14,7 +14,19 @@ const TileGroup = ({ title, titleUrl, description, tiles, compact, intl }) => {
         })}
       >
         <h4 className="coa-TileGroup__title">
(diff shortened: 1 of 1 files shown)


src/components/Tiles/TileGroup.js
```diff
@@ -14,7 +14,19 @@ const TileGroup = ({ title, titleUrl, description, tiles, compact, intl }) => {
...
+                titleUrl.substring(0, 4) === 'http'
+                  ? titleUrl
+                  : `/${intl.locale}${titleUrl}`
+              }
+            >
+              {title}
+            </Link>
+          ) : (
+            title
+          )}
```
briaguya: this is copypasta, and it'd be great to clean this up, but it's a rabbit hole so I made https://github.com/cityofaustin/techstack/issues/3083 instead

PR #550, opened 2019-10-08
get a start on multiple locations

briaguya:
resolves https://github.com/cityofaustin/techstack/issues/3127

### DIFF
--- a/src/components/Contact/ContactDetails.js
+++ b/src/components/Contact/ContactDetails.js
@@ -20,7 +20,18 @@ import {
   phonePropTypes,
 } from './proptypes';
 
-const SingularContactDetails = ({
+const ContactDetails = ({ contacts, intl }) => (
+  <div className="coa-ContactDetails">
+    <SectionHeader isSerif={true}>
+      {intl.formatMessage(i18n.questionsTitle)}
+    </SectionHeader>
+    {contacts.map(c => (
+      <ContactDetailsEntry contact={c} />
+    ))}
+  </div>
+);
+
+const ContactDetailsEntry = ({
   contact: { name, phoneNumber, email, location, hours, socialMedia },
 }) => (
   <React.Fragment>
@@ -38,21 +49,7 @@ const SingularContactDetails = ({
   </React.Fragment>
 );
 
-const ContactDetails = ({ contacts, intl }) => {
-  debugger;
-  return (
-    <div className="coa-ContactDetails">
-      <SectionHeader isSerif={true}>
-        {intl.formatMessage(i18n.questionsTitle)}
-      </SectionHeader>
-      {contacts.map(contact => (
-        <SingularContactDetails contact={contact} />
-      ))}
-    </div>
-  );
-};
-
-SingularContactDetails.propTypes = {
+ContactDetailsEntry.propTypes = {
   contact: PropTypes.shape({
     location: addressPropTypes,
     email: emailPropTypes,
--- a/src/components/Contact/ContactDetails.js
+++ b/src/components/Contact/ContactDetails.js
@@ -52,13 +52,13 @@ const ContactDetails = ({ contacts, intl }) => {
   );
 };
 
-// ContactDetails.propTypes = {
-//   contact: PropTypes.shape({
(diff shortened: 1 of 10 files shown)


src/js/queries/informationPageFragment.js
```diff
@@ -51,6 +51,15 @@ const informationPageFragment = `
           contact {
             name
             email
+            phoneNumber {
```
easherma: Why not use 
https://github.com/cityofaustin/janis/blob/d577f67d0c0f12c810aedb18a2bf56ec55d62532/src/js/queries/phoneFragment.js ?

src/components/Contact/ContactDetails.js
```diff
@@ -20,34 +20,45 @@ import {
...
+  <React.Fragment>
     {/* We want to keep this component easy to read and streamlined, so we are
       avoiding complicated functions in this return.
       Consider breaking into another component and calling that here*/}
-    {email && <Email email={email} />}
+    <h3 className="coa-ContactDetails__name">{name}</h3>
     {location && <Address location={location} />}
-    {phoneNumber && <PhonesList phoneNumbers={phoneNumber} />}
     {hours && !!hours.length && <Hours hours={hours} />}
+    {phoneNumber && <PhonesList phoneNumbers={phoneNumber} />}
```
easherma: did the order get changed in the design?

src/components/Contact/ContactDetails.js
```diff
@@ -20,34 +20,45 @@ import {
...
-    phone: phonePropTypes,
-  }).isRequired,
+const ContactDetails = ({ contacts, intl }) => {
+  debugger;
+  return (
+    <div className="coa-ContactDetails">
+      <SectionHeader isSerif={true}>
+        {intl.formatMessage(i18n.questionsTitle)}
+      </SectionHeader>
+      {contacts.map(contact => (
```
easherma: Might be good to put a comment here pointing out where the contacts get mapped from, took me awhile to find the difference

src/components/Contact/ContactDetails.js
```diff
@@ -20,34 +20,45 @@ import {
...
 );
 
-ContactDetails.propTypes = {
-  contact: PropTypes.shape({
-    location: addressPropTypes,
-    email: emailPropTypes,
-    hours: hoursPropTypes,
-    phone: phonePropTypes,
-  }).isRequired,
+const ContactDetails = ({ contacts, intl }) => {
```
easherma: Some sort of indication of it being plural consistent with the singular might be good. 
ContactDetail ---> ContactDetails
SingularContactDetails ---> AllContactDetails

Or something like that

src/components/Contact/ContactDetails.js
```diff
@@ -20,34 +20,45 @@ import {
...
+        {intl.formatMessage(i18n.questionsTitle)}
+      </SectionHeader>
+      {contacts.map(contact => (
+        <SingularContactDetails contact={contact} />
+      ))}
+    </div>
+  );
 };
 
+// ContactDetails.propTypes = {
```
easherma: is this some outstanding tech debt to make these PropTypes work?

src/components/Contact/ContactDetails.js
briaguya: yes

src/js/queries/informationPageFragment.js
briaguya: i copied from guidepage, I've updated the PR so both info page and guide page use the fragment now

src/components/Contact/ContactDetails.js
briaguya: did some clean up, I think it reads more clearly now

src/components/Contact/ContactDetails.js
briaguya: did some renaming, I think it's a bit clearer now

src/components/Contact/ContactDetails.js
briaguya: Got it uncommented

PR #551, opened 2019-10-08
3095 quick tile fix

briaguya:
fixes https://github.com/cityofaustin/techstack/issues/3095


src/components/Tiles/_Tile.scss
```diff
@@ -34,7 +34,6 @@
   @include shared-tile;
   flex-grow: 1;
   flex-shrink: 1;
-  flex-basis: 0;
```
robbobfrh84: 👍

PR #559, opened 2019-10-10
3152 riffin

briaguya:
Fixes https://github.com/cityofaustin/techstack/issues/3188
Corresponding Joplin PR: https://github.com/cityofaustin/joplin/pull/406
Janis PR app: https://janis-3152-riffin.netlify.com/


src/components/Pages/Guide/GuideSectionCollection.js
```diff
@@ -60,14 +34,7 @@ function GuideSection({
...
       <div className="coa-GuideSection__link">
-        <a
-          href={GetUrlFromTopicsOrDepartments({
-            topics,
-            departments,
-            slug,
-          })}
-          target="_blank"
-        >
+        <a href={url} target="_blank">
```
chiaberry: When is it preferable to use `<Link />` vs `<a />` ? I saw above some refactoring to use Link instead of <a /> and was curious why

src/components/Pages/Departments.js
```diff
@@ -10,23 +10,16 @@ import PageHeader from 'components/PageHeader';
 import SectionHeader from 'components/SectionHeader';
 import TileGroup from 'components/Tiles/TileGroup';
 
-const Departments = ({ departments, intl }) => {
-  const departmentLinks = departments.map(d => ({
-    url: `/${d.slug}`,
-    title: d.title,
-  }));
-
```
chiaberry: does `departments` now come back in this shape from static.config, which is why theres no need to make `departmentLinks` anymore? (does it even come from static.config?)

src/components/Pages/Topic.js
```diff
@@ -43,20 +41,20 @@ const Topic = ({
...
             <TileGroup
-              text={intl.formatMessage(i18n.topServices)}
+              title={intl.formatMessage(i18n.topServices)}
               tiles={topLinks}
             />
           )}
           {!!otherLinks.length && (
             <TileGroup
-              text={intl.formatMessage(i18n.allServices)}
+              title={intl.formatMessage(i18n.allServices)}
```
chiaberry: was this fixing a bug, or is it related to the restructuring? TileGroup doesn't have a text prop in master, so im thinking its a bug fix

src/components/Pages/Guide/GuideSectionCollection.js
briaguya: I've been trying to use <Link> as much as possible so we can utilize all the benefits of the react-static router (https://github.com/react-static/react-static/blob/v5/docs/components.md#link)

I know this is a bit different in newer versions of react-static so it can definitely be an ongoing conversation

src/components/Pages/Departments.js
briaguya: yeah, just realized my comments weren't submitted on janis... sorry about that.

briaguya: https://github.com/cityofaustin/janis/pull/559/files#r337908132

src/components/Pages/Topic.js
briaguya: Yes, if we look at master/prod we'll see that the Topic pages aren't showing "Top services" or "All services" anywhere. This was just a bug find/fix.

static.config.js
```diff
@@ -1,494 +1,609 @@
...
   cleanThemes,
-  cleanServices,
-  cleanInformationPages,
-  clean311,
   cleanNavigation,
-  cleanOfficialDocumentPages,
-  cleanGuidePages,
+  cleanContacts,
+  cleanLinks,
+  cleanDepartmentDirectors,
```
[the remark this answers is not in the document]
chiaberry: is cleanThemes still used anywhere?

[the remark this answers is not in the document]
briaguya: doesn't look like it, good catch, removing it

PR #560, opened 2019-10-11
3065 Guide anchor links needs slowing down—better solution

robbobfrh84:
This is PR will fix that silly "stutter-step" of the scroll effect in our guide pages. Also, it makes sure to scroll to the location if a link to a specific point it followed.

This believe the behavior was being caused by the life-cycle of the url-change-to-react-rendering. 
- When the url is changed, either manually, or with javascript (as we do in this case). the browser "Looks" for the specific `id` right away. HOWEVER, react is saying. no no no, we're gonna rebuild this, AFTER, the fact. That's why I think it jumps. it starts to scroll to the point, but react renders in the middle of the process. 
- I honestly don't know if that is the complete case. But... using `   document.getElementById(anchorTag).scrollIntoView(true);` allows "US" to target the scoll location to our own beat, rather than the build in browser doing it... out of beat. 
- *Check it out! test it! Break it! * 

Fixes this Issue: https://app.zenhub.com/workspaces/techstack-5a78b88e1ce69f3510b678ef/issues/cityofaustin/techstack/3065

Deployed: https://deploy-preview-560--janis.netlify.com/

Before...
Oct-11-2019 13-39-53
-
After...
Oct-11-2019 13-41-18
-
url

### DIFF
--- a/src/components/Pages/Guide/GuideMenu.js
+++ b/src/components/Pages/Guide/GuideMenu.js
@@ -7,8 +7,17 @@ import { isMobileOrTablet } from 'js/helpers/reactMediaQueries';
 function GuideMenuLink({ title, anchorTag, isHeading, isCurrentSection }) {
   // Each GuideSectionWrapper has an id={this.props.anchorTag}
   function goToSection(e) {
-    e.preventDefault();
-    window.location.href = `#${anchorTag}`;
+    // console.log('\n\nclick: go to', anchorTag)
+    // e.preventDefault();
+    // window.location.href = `#${anchorTag}`;
+
+    window.requestAnimationFrame(function(){
+        console.log('\n\nclick: go to', anchorTag)
+        // e.preventDefault();
+        // window.location.href = `#${anchorTag}`;
+        document.getElementById(anchorTag).scrollIntoView(true);
+
+    })
   }
 
   return (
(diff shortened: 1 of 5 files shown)


src/components/Pages/Guide/GuideMenu.js
```diff
@@ -7,8 +7,7 @@ import { isMobileOrTablet } from 'js/helpers/reactMediaQueries';
 function GuideMenuLink({ title, anchorTag, isHeading, isCurrentSection }) {
   // Each GuideSectionWrapper has an id={this.props.anchorTag}
   function goToSection(e) {
-    e.preventDefault();
-    window.location.href = `#${anchorTag}`;
+    document.getElementById(anchorTag).scrollIntoView(true);
```
robbobfrh84: Here is where an element of where the issue was coming from. 
-
At this point, it seems the body of the guide hadn't been updated completely to the DOM, where the scroll-to toward the `id` of the anchor is broken mid path. by targeting the element directly, it'll only start the operation when it's mounted. Rather, then twice through the update of the url.... if that makes sense.

src/components/Recollect/index.js
```diff
@@ -17,6 +17,8 @@ class Recollect extends Component {
       options,
     );
     let loader = new window.Recollect.Widget.Loader(widgetOptions);
+    const scrollToId = window.location.hash.split('#')[1]
+    document.getElementById(scrollToId).scrollIntoView(true);
```
robbobfrh84: ..same here for auto scrolling when a link is clicked to this location. Copy/paste a link into a new tab to see the effect!

src/components/Pages/Guide/GuideMenu.js
Niicck: Awesome!

PR #565, opened 2019-10-14
3092 Janis patches for guide with missing page.

robbobfrh84:
# Description

This PR will fixes Janis builds that broke when guide pages published without pages. 

Fixes # (issue)

* [Issue Link](https://app.zenhub.com/workspaces/techstack-5a78b88e1ce69f3510b678ef/issues/cityofaustin/techstack/3092)

# Testing Notes

To Reproduce

- Go to joplin
- Make a guide page
- Add a section without a page
  - Now, if you attempt to publish, should should be notified that you need to add a page. 
- So, add a page and publish the test guide. 
- Then, Delete that page!
- Janis should now still build and show the page-less guide without breaking!

* [Deployed Link]()

# Checklist:
- [x] Request reviewers
- [x] Slack-out message for review

### DIFF
--- a/src/components/Pages/Guide/GuideSectionCollection.js
+++ b/src/components/Pages/Guide/GuideSectionCollection.js
@@ -38,13 +38,18 @@ function GuideSection({
   sectionHeading,
   intl,
 }) {
-  const pageData = page.informationPage || page.servicePage;
+  console.log("page :", page)
+  const pageData = page.informationPage || page.servicePage || {};
+  console.log("pageData :", pageData)
   const title = pageData.title;
   const description = page.informationPage
     ? pageData.description
     : pageData.shortDescription;
   const additionalContent = pageData.additionalContent;
+  // const topics = pageData.topics.edges;
   const topics = pageData.topics.edges;
+  console.log("topics :", topics)
+
   const departments = pageData.relatedDepartments.edges;
   const slug = pageData.slug;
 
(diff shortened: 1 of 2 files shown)


src/components/Recollect/index.js
```diff
@@ -18,7 +18,9 @@ class Recollect extends Component {
     );
     let loader = new window.Recollect.Widget.Loader(widgetOptions);
     const scrollToId = window.location.hash.split('#')[1];
-    document.getElementById(scrollToId).scrollIntoView(true);
```
robbobfrh84: I caught this additional bug while doing this fix. Sometimes, when going to a guide page *without* a specific scroll-to point in the guide, you'll get an error `cannot scrollTo of undefined`. This catches this case when it draw the error and won't auto-scroll.

src/components/Pages/Guide/GuideSectionCollection.js
```diff
@@ -101,13 +101,18 @@ function GuideSectionCollection({
...
           resizeCount={resizeCount}
         >
-          <GuideSection
-            page={page}
-            pageNumber={index + 1}
-            numberOfPages={section.pages.length}
-            sectionHeading={section.heading}
-            intl={intl}
-          />
+        { (page.servicePage || page.informationPage) &&
```
robbobfrh84: Here, we'll break adding the content for the page if it dosen't have either a servicePage or an information page. This will keep pageless guide sections from breaking when built.

PR #600, opened 2019-11-07
3205 - only hide scrollbar for guide TOC, show all others

briaguya:
# Description

make it so scrollbars show up everywhere EXCEPT the guide page TOC (instead of hiding them everywhere)

Fixes:

* Issue Link https://github.com/cityofaustin/techstack/issues/3205

# Testing Notes

make sure scrollbars show up in all sorts of different browsers.

* Deployed Link: https://janis-3205-scrollbars.netlify.com/en/health-safety/healthcare-prevention/disease-prevention/mobile-food-vendor-permit-guide

# Checklist:
- [x] Request reviewers
- [x] Link this PR in the issue

### DIFF
--- a/src/components/Pages/Guide/_Guide.scss
+++ b/src/components/Pages/Guide/_Guide.scss
@@ -80,13 +80,17 @@ $max-guide-tablet-width: $coa-medium-screen;
   margin-top: $sticky-guide-top;
   top: 0px;
 
-  // Hide the nav scrollbar in Chrome/Safari
+  // Hide TOC scrollbar in:
+  // Webkit eg. Chrome/Safari
   &::-webkit-scrollbar {
     width: 0 !important;
   }
 
-  // Hide the nav scrollbar in edge/ie
+  // Edge/Internet Explorer
   -ms-overflow-style: none;
+
+  // Firefox
+  scrollbar-width: none;
 }
 
 .coa-GuidePage__main-content {
--- a/src/components/Pages/Guide/_Guide.scss
+++ b/src/components/Pages/Guide/_Guide.scss
@@ -79,6 +79,14 @@ $max-guide-tablet-width: $coa-medium-screen;
   overflow-y: scroll;
   margin-top: $sticky-guide-top;
   top: 0px;
+
+  // Hide the nav scrollbar in Chrome/Safari
+  &::-webkit-scrollbar {
+    width: 0 !important;
+  }
+
+  // Hide the nav scrollbar in edge/ie
+  -ms-overflow-style: none;
 }
 
 .coa-GuidePage__main-content {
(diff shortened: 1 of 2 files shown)


src/components/Pages/Guide/_Guide.scss
```diff
@@ -79,6 +79,14 @@ $max-guide-tablet-width: $coa-medium-screen;
   overflow-y: scroll;
   margin-top: $sticky-guide-top;
   top: 0px;
+
+  // Hide the nav scrollbar in Chrome/Safari
```
easherma: Do we know which one of these applies to Firefox? I tested quickly and didn't see a scrollbar on the guide TOC so I think we're good. 

Also can we update the note to say something like:
" Hides the nav scrollbar for guide page (used for the Guide Menu) "

Better yet could we add a class like coa-GuidePage__content-container--hide-scrollbar so its a bit more self-documenting?

briaguya: It's in the `.coa-GuideMenu__desktop-container` class, I'm not sure how (used for guide menu) makes that more clear. I feel like adding another class for it might get a little too verbose for my liking.

That's interesting that it's working in firefox, I forgot to check that yesterday. I saw this: https://stackoverflow.com/a/53787170 but haven't actually put that in there yet. I'll poke at it a bit

easherma: Yeah! I saw that too and just figured there was a reason why it was the way it was. 

What I was thinking was ```.coa-GuideMenu__desktop-container--hide-scrollbar``` would follow BEM convention and also be more self-documenting, so we probably wouldn't need a verbose comment as well. 

But I'm far from an expert on BEM or css stuff, so I defer to the team overall.

Niicck: I'm a fan of having explicit comments within classes. That isn't something to be avoided. `.coa-GuideMenu__desktop-container` is verbose enough to let us know that this class is for containers on desktop on a guide page. If a style only needs to be added to one class, it should not be separated out into separate classes. Even `.coa-GuideMenu__desktop-container--hide-scrollbar` isn't self-documenting enough. Following that logic, we'd have to break it down into `.coa-GuideMenu__desktop-container--hide-scrollbar-chrome-and-safari`. 

Comments in css are useful, it's good to know why styles are added. And reducing complexity and bloat by keeping things in as few classes as possible is preferred.

src/components/Pages/Guide/_Guide.scss
```diff
@@ -79,6 +79,18 @@ $max-guide-tablet-width: $coa-medium-screen;
...
+  // Webkit eg. Chrome/Safari
+  &::-webkit-scrollbar {
+    width: 0 !important;
+  }
+
+  // Edge/Internet Explorer
+  -ms-overflow-style: none;
+
+  // Firefox
+  scrollbar-width: none;
```
briaguya: tested in FF and found this was definitely the thing making it work. There was probably some scss magic happening before to make it work but now it's explicit.

PR #604, opened 2019-11-08
3219 en dash contacts

chiaberry:
# Description

I changed the hyphen in the hours to an en dash "–" instead of the hyphen it had "-"

I updated spacing so that "5 am–3 pm", space between the time and the pm/am but no space to the en dash. 

*also* the style guide says to use Noon instead of 12 pm. I could not find in the moment docs an automatic format to add so instead if the time is '12:00:00' return 'Noon'. 

I also updated some proptypes and added a key to get rid of react warnings that were popping up.

# Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

# How can this be tested?

Go here https://janis-3219-en-dash-contacts.netlify.com/en/police-oversight/: to see the hours listed. The dash is a en dash instead of a hyphen. En dashes are slightly longer than hyphens but shorter than em dashes. You can highlight the punctuation and google search it to confirm its an en dash. 

 

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/src/components/Contact/ContactDetails.js
+++ b/src/components/Contact/ContactDetails.js
@@ -25,8 +25,8 @@ const ContactDetails = ({ contacts, intl }) => (
     <SectionHeader isSerif={true}>
       {intl.formatMessage(i18n.questionsTitle)}
     </SectionHeader>
-    {contacts.map(c => (
-      <ContactDetailsEntry contact={c} />
+    {contacts.map((c, index) => (
+      <ContactDetailsEntry contact={c} key={index} />
     ))}
   </div>
 );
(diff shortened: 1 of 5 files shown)


src/components/Contact/ContactDetails.js
```diff
@@ -25,8 +25,8 @@ const ContactDetails = ({ contacts, intl }) => (
     <SectionHeader isSerif={true}>
       {intl.formatMessage(i18n.questionsTitle)}
     </SectionHeader>
-    {contacts.map(c => (
-      <ContactDetailsEntry contact={c} />
+    {contacts.map((c, index) => (
```
robbobfrh84: 👍

PR #607, opened 2019-11-12
adding dynamic content back, lets see if it causes any problems

easherma:
# Description

# Types of changes

# How can this be tested?

 
Dyanmic content, if selected in Joplin, should render on the page, not nested inside a step. 
https://janis-3366-static-map.netlify.com/en/housing-utilities/recycling-trash-and-compost/household-waste/hazardous-waste-dropoff/

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/src/components/Pages/Service.js
+++ b/src/components/Pages/Service.js
@@ -24,6 +24,7 @@ const Service = ({ service, intl }) => {
       slug,
       steps,
       additionalContent,
+      dynamicContent,
       contacts,
       related,
       shortDescription,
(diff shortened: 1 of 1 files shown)


src/components/Pages/Service.js
```diff
@@ -68,26 +69,31 @@ const Service = ({ service, intl }) => {
                     )
                   ) : null}
 
+                  {!!dynamicContent &&
+                    dynamicContent.map(content => (
```
robbobfrh84: What is the significance of the `!!` here? The way I understand it, `!!dynamicContent == dynamicContent`

Niicck: Would we also want to check that dynamicContent has a "length" like we do for contacts?

easherma: The short answer is that this is just adding back code that already was here before and worked. 

Longer answer; it's a way of checking if the object exists:
https://medium.com/better-programming/javascript-bang-bang-i-shot-you-down-use-of-double-bangs-in-javascript-7c9d94446054

PR #608, opened 2019-11-12
3248 set up and explore testing frameworks

chiaberry:
# Description

I've set up **Jest** to do snapshot tests for components. After you run `test components`, it will compare the rendering to what is stored. This is rendered with dummy data / static data, so the only thing that is being tested is that the json structure of the component matches what was run previously. A test failing means that the structure of the component has changed in some way. The snapshots will check if class names are changed but *not* if the css was changed in the class name. If the changes are expected, update the snapshot.

I have also set up **Cypress**, which can run integration tests. I've written a test to check that the site loads as well as a test that checks if recollect is loaded and that the OPO document links contain the url for the s3 bucket.

I updated the existing storybook components so they work, except for the Footer. The HOC wrapping throws an error and I didn't think it was important enough to spend more time on it. Let me know if anyone disagrees.  

I've also done some general cleanup, proptypes etc. I also moved the themes navigation data out of the stories folder and reduced it to an array of themes. 

https://github.com/cityofaustin/techstack/issues/3248

# Types of changes

# How can this be tested?
`yarn test` runs the jest test suites
`yarn open-cypress` launches the cypress test runner. You should have a local instance of janis running before launching cypress

# Checklist:
- [x] Request reviewers
- [x] Slack-out message for review
- [x] Link this PR in the issue


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
chiaberry: I tried deleting all these cypress boilerplate/sample files and every time i run the test runner, they are added. womp womp.

src/components/PageSections/ContextualNav/index.js
```diff
@@ -22,7 +23,7 @@ const ContextualNav = ({ parent, relatedTo, offeredBy, intl }) => (
               {`${intl.formatMessage(i18n.relatedTo)}: `}
             </span>
             {relatedTo.map((relatedLinkData, index) => (
-              <Link to={`/${intl.locale}${relatedLinkData.url}`}>
+              <Link to={`/${intl.locale}${relatedLinkData.url}`} key={relatedLinkData.url}>
```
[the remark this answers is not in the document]
Niicck: The `index` alone would also work as a key, which is a common way people handle this in react. But the unique url also works.

cypress/fixtures/example.json
Niicck: Can they be .gitignored?

src/components/PageSections/Menu/ThemesNav.js
```diff
@@ -3,13 +3,52 @@ import { injectIntl } from 'react-intl';
...
+    'التصاريح والتذاكر',
+    'الإسكان والمنافع',
+    'لحيوانات الأليفة',
+    'الصحة والسلامة',
+    'الاستكشاف والزيارة',
+    'الحكومة والأعمال',
+    'الوظائف'
+  ],
+};
+
```
Niicck: Should this be placed with the rest of our locales stuff?

src/components/Pages/Service.js
```diff
@@ -66,11 +66,10 @@ const Service = ({ service, intl }) => {
                         <Steps steps={steps} />
                       </Fragment>
                     )
-                  ) : null}
+                  )}
```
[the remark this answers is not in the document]
robbobfrh84: 👍? 👍: 👍

package.json
```diff
@@ -42,23 +42,29 @@
     "react-router-dom": "^5.1.2",
     "react-static": "^7.2.2",
     "react-static-plugin-react-router": "^7.2.2",
+    "react-test-renderer": "^16.11.0",
```
Niicck: Probably this and storybook are dev dependencies

src/components/PageSections/ContextualNav/index.js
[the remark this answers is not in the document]
chiaberry: index as key is technically an antipattern, but since these components aren't going to be reordered, you're right that the key would have been just fine

cypress/fixtures/example.json
chiaberry: oh what a good idea! updating now.

package.json
chiaberry: right again. updating now

PR #623, opened 2019-11-24
Issue 3297: location MVP Janis

Niicck:
https://github.com/cityofaustin/techstack/issues/3297

# Description

We've done it. An MVP for the LocationPage on Janis. It works with dummy data in storybook, but not real Joplin data yet.

Bonus things happened:
- [ ] We upgraded storybook! I used a CLI script to automagically migrate all of our past storybook stories to the new version.
- [ ] We upgraded react-intl! We can now import `intl` using hooks instead of `injectIntl`, which is wonderful.
- [ ] We removed componentWillReceiveProps from I18nController!

# Types of changes

# How can this be tested?

Since there are no Joplin Location Pages yet, you can test it in storybook!
https://janis-location-storybook.netlify.com/?path=/story/pages-locationpage--basic

To prove that my changes don't break the original site: 
https://janis-3297-location.netlify.com/

# Checklist:
- [ ] Request reviewers


src/js/helpers/constants.js
```diff
@@ -15,6 +15,16 @@ export const WEEKDAY_MAP = {
   SATURDAY: 6,
 };
 
-export const FULL_WIDTH_RESPONSIVE_IMAGE_SIZES = 
+export const DAYS = () => ([
```
chiaberry: why a function and not an array?

Niicck: Excellent question! Maybe a strange design choice that we wouldn't want to adopt, but there is a reason for it.

When we import constants, they are mutable. If you require a constant and change it `DAYS.push(x), DAYS.splice(y)` that will actually change the original reference. So if another file or process requires that file, they'll receive the mutated version, not the original `DAYS`.

This was a tricky bug that I ran into in my days as a young developer.

So if we return a distinct function that returns an array, every instance of `import DAYS from ...; const days = DAYS();` will return a distinct object.

The alternative would be to remember to destruture DAYS on every invocation `{...DAYS}` to make a copy. But someone forgets to do that, then something will break.

It's a safeguard to make our constant work as expected. I'm open to other suggestions about it.

chiaberry: makes perfect sense to me. thanks for explaining it so well

PR #651, opened 2019-12-17
Fix build error

briaguya:


static.config.js
```diff
@@ -376,11 +377,12 @@ const getInformationPageData = async (
     client,
   );
 
-  if (pagesOfGuides[id]) { // We're checking if this id is part of guide page because it may not be published and draw and error.
-    informationPage.pageIsPartOf = pagesOfGuides[id]
+  if (pagesOfGuides && pagesOfGuides[id]) {
```
briaguya: this is the line the error was coming from, because when pagesOfGuides is null we can't try to index in

PR #674, opened 2020-01-16
pull in exceptions, also fix colon bug

easherma:
# Description

# Types of changes

# How can this be tested?

 
     

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/src/components/Pages/Location/LocationInfo.js
+++ b/src/components/Pages/Location/LocationInfo.js
@@ -26,7 +26,7 @@ const LocationPageContact = ({ phone, email }) => {
   // TODO: use 'libphonenumber-js' for real phone from Joplin API
   const phoneContent = (
     <Fragment>
-      {`${phone.name}: `}
+      {!!phone.name && `${phone.name}: `}
       <a href={`tel:${phone.value}`}>{phone.value}</a>
     </Fragment>
   );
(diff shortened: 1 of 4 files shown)


src/components/Pages/Location/LocationInfo.js
```diff
@@ -26,7 +26,7 @@ const LocationPageContact = ({ phone, email }) => {
   // TODO: use 'libphonenumber-js' for real phone from Joplin API
   const phoneContent = (
     <Fragment>
-      {`${phone.name}: `}
+      {!!phone.name && `${phone.name}: `}
```
easherma: Fixed a bug where the colon would show if there was no phone 'name'

@Niicck is there any particular reason why we didn't just re-use the Phone component in Contact? We had fixed this bug previously before there.

Niicck: A couple reasons.

1. I didn't have real Joplin data to work with when I wrote it. So the Phone component wouldn't have parsed correctly. I made a note to use `// TODO: use 'libphonenumber-js' for real phone from Joplin API` when the Joplin API got hooked up.
2. The existing Phone component is not a "Phone" component, but a "Phones" component. It has a material icon and different formatting than the design specs for the Location phoneContent section.
  - A valid option would be to abstract out an individual "Phone" component that could be used within both "Phones" and "Location". I did not choose to do that. Instead I wrote 4 lines of code to get the location phone section working and get the mvp shipped. But writing a "Phone" component is a wonderful valid option to improve code consistency and prevent bugs like the one you just fixed.

easherma: Cool, so we should probably make an issue to do that and evaluate what to do for the components in upcoming events

PR #682, opened 2020-01-30
3840 Locations hours display

robbobfrh84:
Zenhub Link: https://app.zenhub.com/workspaces/techstack-5a78b88e1ce69f3510b678ef/issues/cityofaustin/techstack/3840

# Description

This PR will add design and width brackdown for your locations components. See ☝️issue link above for design. 

@briaguya and @Niicck you both built this out initially - So, I did my best to stick with our html and css patter here. However, you may catch something that could be organized better. 

# Types of changes

# How can this be tested?

 
Deployed: https://janis-3840-location-hours.netlify.com/
- Mixed Example:  https://janis-3840-location-hours.netlify.com/en/location/st-john-neighborhood-center/
- Simple Example:  https://janis-3840-location-hours.netlify.com/en/location/recycle-reuse-drop-off-center/

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/src/components/Pages/Location/LocationInfo.js
+++ b/src/components/Pages/Location/LocationInfo.js
@@ -148,15 +148,10 @@ const LocationPageFacilityHours = ({ hours }) => {
     </div>
   );
 
-  // ⏰ ⏰ ⏰ ⏰ ⏰ ⏰
-  // ⏰ ⏰ ⏰ ⏰ ⏰ ⏰
-  // ⏰ ⏰ ⏰ ⏰ ⏰ ⏰
-  // ⏰ ⏰ ⏰ ⏰ ⏰ ⏰
-
   return (
     <div className="coa-LocationPage__sub-section">
       <h2 className="coa-LocationPage__sub-section-title">
-        ⏰{intl.formatMessage(i18nLocations.facilityHours)}
+        {intl.formatMessage(i18nLocations.facilityHours)}
       </h2>
       <div className="coa-LocationPage__sub-section-block-container">
         <LocationPageBlock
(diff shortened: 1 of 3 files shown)


src/components/Pages/Location/LocationInfo.js
```diff
@@ -131,7 +131,14 @@ const LocationPageFacilityHours = ({ hours }) => {
...
                 {hours[day] !== null
-                  ? hours[day]
+                  ? hours[day].split(',').map( (dateRange, i) => {
+                    const comma = i+1 < hours[day].split(',').length ? "," : ""
+                    return (
+                      <div class="coa-LocationPage__hours-section">
+                        {dateRange}{comma}&nbsp;
+                      </div>
+                    )
+                  })
```
robbobfrh84: This was a bit tricky. What we're doing here is adding a comma `,` AND additional `<div>` for dates with multiple ranges. I racked my head on a clearer way to do this, but wasn't finding the muse🧚‍♂️.

src/components/Pages/Location/LocationInfo.js
```diff
@@ -152,10 +159,14 @@ const LocationPageFacilityHours = ({ hours }) => {
...
-            content={hours.exceptions}
-          />
+          <div className="coa-LocationPage__sub-section-block">
+            <div className="coa-LocationPage__sub-section-block-title-padded">
+              {intl.formatMessage(i18nContact.exceptions)}
+            </div>
+            <div className="coa-LocationPage__sub-section-block-contents">
+              {hours.exceptions}
+            </div>
+          </div>
```
robbobfrh84: The `LocationPageBlock` was a bit too restrictive for what we needed to do here. So, rather than put some conditionals there, i just modified it here as part of the specific component.

src/components/Pages/Location/LocationInfo.js
Niicck: It took me a second, but I get what's going on here. Because if we're not the last dateRange, then we need to add a comma. I think it works.

PR #684, opened 2020-02-04
3740 service hours from location

easherma:
# Description

We ended up needing to do some dancing around with logic to meet MVP acceptance criteria for #3740

I moved much of this conditional logic into what felt like the appropriate components, though it all feels a little messy. There aren't quite any clear lines of where we'd delineate responsibilities by component. I agree with @briaguya that a better longer term solution would probably be to refactor the backend API to handle the logic and send the appropriate chunk of hours over. 

https://github.com/cityofaustin/joplin/pull/536
I also pushed up a version where we actually run the validation to prevent authors from sending an ambigious case over to janis, which if merged could simplify the logic here. 
https://github.com/cityofaustin/joplin/pull/551
# Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
(we're adding a new data field so Travis is currently unhappy)

# How can this be tested?

 
     
See issue for testing links. 

It would be great to have some unit tests for this code, but as we've discussed previously the current architecture of components in janis makes it pretty hard to know where/how to start writing useful tests. 

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/src/components/Pages/Location/LocationServiceList.js
+++ b/src/components/Pages/Location/LocationServiceList.js
@@ -24,7 +24,6 @@ const DayHours = ({ day, hours }) => {
   );
 };
 
-
 const HoursExceptions = ({
   serviceHoursExceptions,
   locationHoursExceptions,
@@ -49,8 +48,18 @@ const ServiceHours = ({ serviceHours, locationHours, hoursSameAsLocation }) => {
   const [expanded, setExpanded] = useState(false);
   const intl = useIntl();
   let hours = serviceHours;
+  let hoursAreEmpty = Object.values(serviceHours).every(
+    entry => entry == null || entry == undefined || entry == '',
+  );
   if (hoursSameAsLocation) {
     hours = locationHours;
+    hoursAreEmpty = Object.values(locationHours).every(
+      entry => entry == null || entry == undefined || entry == '',
+    );
+  }
+  let showHours = true;
+  if (!hoursSameAsLocation && hoursAreEmpty) {
+    showHours = false;
   }
 
   const days = getDaysInOrder();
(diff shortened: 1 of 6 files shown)


src/components/Pages/Location/LocationServiceList.js
```diff
@@ -24,26 +24,43 @@ const DayHours = ({ day, hours }) => {
...
   return (
     <div>
       <div className="coa-ContactHoursExceptionsTitle">
-        {intl.formatMessage(i18nContact.exceptions)}
-      </div>
-      <div className="coa-ContactHoursExceptions">
-        {exceptions}
+        {!!exceptions && (intl.formatMessage(i18nContact.exceptions))}
       </div>
+      <div className="coa-ContactHoursExceptions">{exceptions}</div>
```
chiaberry: if there are no exceptions, the message will not show up but does the empty div still appear?

src/js/helpers/cleanData.js
```diff
@@ -85,6 +85,14 @@ export const cleanLocationPageJanisUrl = janisUrl => {
...
 };
 
+/*
+tech debt: this code is used for service hours as well, but the variable defition
+implies otherwise, could be a source of bugs
+
+also the best place to shape the data is probably the backend, why do we have to
+transform the data from the backend so much here? If so shouldn't we reformat the
+data on the backend instead, either in the model or graphql schema?
+*/
```
chiaberry: 👍

src/components/Pages/Location/LocationServiceList.js
chiaberry: oh maybe we need the empty div to be there for spacing reasons?

easherma: good question... !

PR #685, opened 2020-02-05
Window build fix error

robbobfrh84:
This fix looks at the recent build fails in our pipeline. Once again, I've re-discovered that `window` during builds doesn't exist, so it fails. 

But, that's ok, because react hooks can come to the rescue!

I figured out a way to use hooks without a variable. Pretty much, this will act as an event that fires when the component has been added to the dom. allowing you to do fun stuff with js!

```javascript
  useEffect(() => {
    // you can now do stuff with your components html!
  }, [])
```

### DIFF
--- a/src/components/Pages/Form.js
+++ b/src/components/Pages/Form.js
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react';
+import React, { useRef, useEffect } from 'react';
 import { withRouteData, Head } from 'react-static';
 import { injectIntl } from 'react-intl';
 /**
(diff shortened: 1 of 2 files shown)


src/components/Pages/Form.js
```diff
@@ -60,13 +60,14 @@ function FormContainer({
...
-    style: "popup"
-  })
+  useEffect(() => {
+    loader.start({
+      contentId: 'coa-iFrameForm',
+      loaderId: 'coa-loadingWheel',
+      errorId: 'coa-LockedIframeRequestMessage',
+      style: "popup"
+    })
+  }, []) // By having the second argument as an empty string, it will act as "ComponentDidMount" and execute once the component's elements have been added to the dom.
```
chiaberry: that's really neat

PR #689, opened 2020-02-10
Looks like these have all the changes we need?

easherma:
# Description
Related to https://github.com/cityofaustin/joplin/pull/555 
That joplin fix stripped out the custom resolver for service pages, so the GraphQL query needed is ever so slightly different; basically steps doesn't need (and cant have) a sub selection, but the fields are defined as part of that object anyways. But it means we need to tweak where that data is referenced. 

# Types of changes

# How can this be tested?

 
     

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/src/components/Steps/index.js
+++ b/src/components/Steps/index.js
@@ -9,7 +9,7 @@ import { stepsPropTypes } from './proptypes';
 const mapSteps = (steps, singleStep) =>
   // singleStep boolean param to determine whether to render steps in a list (ul) or a single item for StepWithOptions and StepBasic
   steps.map((step, index) => {
-    if (step.stepType === 'step_with_options_accordian')
+    if (step.type === 'step_with_options_accordian')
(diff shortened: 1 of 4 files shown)


src/components/Steps/index.js
```diff
@@ -9,7 +9,7 @@ import { stepsPropTypes } from './proptypes';
 const mapSteps = (steps, singleStep) =>
   // singleStep boolean param to determine whether to render steps in a list (ul) or a single item for StepWithOptions and StepBasic
   steps.map((step, index) => {
-    if (step.stepType === 'step_with_options_accordian')
+    if (step.type === 'step_with_options_accordian')
```
briaguya: just curious, when/where did this change?

easherma: Its affected on the joplin side, since essentially its just passing on the name of the model field

PR #695, opened 2020-02-18
3817 event lists view

chiaberry:
# Description

The Events List view. I've put comments in line, but the gist is that I've added the new Event List View, copying how the Official Documents Pagination is set up. Event List > EventListPagination  which shows EventListEntries. I moved some of the functions and components out of Official Documents, so both documents and events can use them.

There is a joplin PR out there to adjust the query so I can get the events in date order, after "today". 

Clicking the Event Entry links to the Event Page. 

I had to handle some localization in cleanData.js (Noon -> Mediodia), let me know if that makes sense how I did it. 

If you are going to test/run locally, you should be sure to use `yarn start-joplin-pr` because otherwise the events query will fail on staging/production

[Joplin Related Pull Request Link](https://github.com/cityofaustin/joplin/pull/563)

# Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How can this be tested?

 
     

https://janis-3817-event-lists.netlify.com/events/

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/src/components/PageSections/Pagination/index.js
+++ b/src/components/PageSections/Pagination/index.js
@@ -1,0 +1,21 @@
+import React from 'react';
+
+export const PageNumber = ({ pageNumber, index, changePage, paginationIndex, pageNumberIndex, contentType }) => {
+  const active = pageNumber === index ? " active" : ''
+  let ellipsis = ""
+  let pageIndex = index
+  if (index === "...") {
+    ellipsis = " ellipsis"
+    pageIndex = pageNumberIndex > paginationIndex ? pageNumber-1 : pageNumber+1
+  }
+  return (
+    <div
+      onClick={()=>changePage(pageIndex-1)}
+      className={ `coa-${contentType}_page number-container` + active + ellipsis }
+    >
+      <div className={ `coa-${contentType}_number` }>
+        { index }
+      </div>
+    </div>
+  )
+}
--- a/src/components/Pages/EventList/EventListEntry.js
+++ b/src/components/Pages/EventList/EventListEntry.js
@@ -1,0 +1,101 @@
+import React from 'react';
+import moment from 'moment-timezone';
+import { useIntl } from 'react-intl';
+import { Link } from 'react-router-dom';
+
+import { useMobileQuery } from 'js/helpers/reactMediaQueries.js';
+import { formatTimeLang } from 'js/helpers/cleanData';
+import { events as i18n } from 'js/i18n/definitions';
+
+const EventDateCalendar = ({date}) => {
+  const intl = useIntl();
+  const isMobile = useMobileQuery()
+  
+  moment.locale(intl.locale);
+  let monthType = isMobile ? 'MMM' : 'MMMM';
+  let momentMonth = moment(date, 'YYYY-MM-DD').format(monthType);
+  // Spanish is adding periods at the end of abbreviations on mobile
+  if (momentMonth.slice(-1) === '.') {
+    momentMonth = momentMonth.slice(0, -1)
+  }
+  // Spanish also lowercases the months
+  momentMonth = momentMonth.charAt(0).toUpperCase() + momentMonth.slice(1);
+  // the Date is the last two digits of the date
+  let momentDate = date && date.slice(-2)
+
+  return (
+    <div className="coa-EventListPage__DateCal">
+      <div className="coa-EventListPage__DateCal-month">
+        {momentMonth}
+      </div>
+      <div className="coa-EventListPage__DateCal-date">
+        {momentDate}
+      </div>
+    </div>
+  )
+}
+
+
+const EventDateListDetails = ({ event }) => {
+  const intl = useIntl();
+  const isMobile = useMobileQuery()
+  let dayType = isMobile ? 'ddd' : 'dddd';
+  const { date, startTime, endTime, location, title, eventIsFree, registrationUrl, eventUrl, canceled, feesRange } = event;
+  const noon = intl.formatMessage(i18n.noon);
+
+  // parse Date
+  moment.locale(intl.locale)
+  let momentDay = moment(date, 'YYYY-MM-DD').format(dayType)
+  momentDay = momentDay.charAt(0).toUpperCase() + momentDay.slice(1);
+  if (momentDay.slice(-1) === '.') {
+    momentDay = momentDay.slice(0, -1)
+  }
+
+  let eventTime = startTime && endTime
+    ? `${formatTimeLang(startTime, noon)}–${formatTimeLang(endTime, noon)}`
+    : formatTimeLang(startTime, noon);
+
+  let locationName = null
+  if (location) {
+    locationName = (location.locationType === 'city_location')
+      ? location.cityLocation.title
+      : location.remoteLocation.name;
+  }
+
+  // Joplin lets a user mark an event as free but also include costs.
+  // Or not mark as free, but not include fees. Ideally Id like to just check if eventIsFree
+  // let cost = (eventIsFree) ? `${intl.formatMessage(i18n.free)}` : feesRange;
+  // In the interim, this will have to do. b/c if a user doesnt enter fees, it should say free -chia
+  let cost = (feesRange.length === 0) ? `${intl.formatMessage(i18n.free)}` : feesRange;
+  let registration = registrationUrl.length ? `• ${intl.formatMessage(i18n.registrationReq)}`: '';
+
+  if (cost.slice(0,2) === '$0') {
+    cost = `${intl.formatMessage(i18n.free)}${cost.slice(2)}`
+  }
+
+  return (
+    <div className="coa-EventListPage__EntryDetails"> 
+      { 
+        canceled 
+        && <div className="coa-EventListPage__Canceled">
+            {intl.formatMessage(i18n.canceled)}
+           </div> 
+      }
+      <div className="coa-EventListPage__Time">{ `${momentDay} • ${eventTime}`}</div>
+      <div className="coa-EventListPage__Title"><a href={eventUrl}>{title}</a></div> 
+      <div className="coa-EventListPage__Location">{locationName}</div>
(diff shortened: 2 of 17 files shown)


src/components/Pages/EventList/index.js
```diff
@@ -0,0 +1,36 @@
...
+import EventListPagination from 'components/Pages/EventList/EventListPagination';
+import UserFeedback from 'components/UserFeedback';
+import { events as i18n } from 'js/i18n/definitions';
+
+const filterOnDate = event => {
+  // the api returns events greater than or equal to the time when the site is built
+  // this filters out the ones between the build time and now
+  const dateNow = moment().format('YYYY-MM-DD')
+  return moment(event.date).isSameOrAfter(dateNow);
+}
```
chiaberry: Should I limit this to be the current time in Austin? Or filter it based on wherever the user is? I am thinking Austin, but wanna hear feedback

src/js/helpers/cleanData.js
```diff
@@ -2,13 +2,15 @@ import { findKey } from 'lodash';
...
 import { parsePhoneNumberFromString } from 'libphonenumber-js';
 
 import { WEEKDAY_MAP } from 'js/helpers/constants';
 
 export const formatTime = time => {
+  let lang = Cookies.get('lang')
   if (time === '12:00:00') {
-    return 'Noon';
+    return lang === 'es' ? 'Mediodía' : 'Noon';
   }
```
chiaberry: is there a better way to handle the localization outside of React components? This obviously doesn't handle our other languages, but I figured for MVP....

src/js/helpers/pagination.js
```diff
@@ -0,0 +1,26 @@
+export function buildPages(allData, itemsPerPage) {
+  const pages = []
```
chiaberry: the functions from Official Documents Pagination, moved out

static.config.js
```diff
@@ -569,6 +572,33 @@ const getEventPageData = async (id, client) => {
   return { eventPage: eventPage };
 };
 
+const getAllEvents = async client => {
+  // Is this how we want to handle it? 
+  const date_now = moment().format('YYYY-MM-DD')
```
chiaberry: considering this also being toggled as Austin timezone, im trying to figure out where netlify's timezone is

static.config.js
```diff
@@ -903,10 +947,10 @@ export default {
...
     const routes = [
-      {
-        path: '/search',
-        template: 'src/components/Pages/Search', //TODO: update search page to be conscious of all languages
-      },
+      // {
+      //   path: '/search',
+      //   template: 'src/components/Pages/Search', //TODO: update search page to be conscious of all languages
+      // },
       {
```
chiaberry: nothing to see here

src/components/Pages/OfficialDocuments/OfficialDocumentPage.js
```diff
@@ -103,42 +76,22 @@ const OfficialDocumentPage = ({ officialDocuments, intl }) => {
...
-              <UserFeedback />
+        <UserFeedback />
       </div>
     </div>
   )
 
 }
 
-const PageNumber = ({ pageNumber, index, changePage, paginationIndex, pageNumberIndex })=>{
-  const active = pageNumber === index ? " active" : ''
```
chiaberry: moved to component

src/components/Pages/EventList/index.js
robbobfrh84: This is a good question. I think content would have a good point of view. I'd lean keeping it local time with CST added. for clarification. Like, if i'm in another time zone, i'm already thinking about that and to see CST i know for sure. I'm often more confused when it's updated to my current location while traveling. But, maybe good for this iteration and a thought down the road...

src/js/helpers/cleanData.js
robbobfrh84: You can use the `.json`s we have for translations in `https://github.com/cityofaustin/janis/blob/master/src/js/i18n/locales`

static.config.js
robbobfrh84: you can just force it always be CST
- something like this
- `moment.utc('07-18-2013 16:10:11', 'MM-DD-YYYY HH:mm').tz('America/Chicago').format("YYYY-MM-DD HH:mm");
`
- I don't think there is an Austin specific as the timezone is anchored to Chicago i guess. (Why does Chicago think it's so special!) But, that accounts for daylight saving too, which is good!

static.config.js
robbobfrh84: 🙈

src/components/Pages/OfficialDocuments/OfficialDocumentPage.js
robbobfrh84: thumb👍

src/components/PageSections/Pagination/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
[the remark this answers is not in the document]
robbobfrh84: 👏

src/components/Pages/EventList/index.js
Niicck: That's a good content question. It might be confusing to adjust the eventList based on the user's location because we're not adjusting the individual event datetimes based on the user's location. That would be a bad user experience. Someone from New York looks at the site, or is planning a trip schedule, whatever. "Oh there's an event at 5:00 on the 26th." Intuitively they'd expect that to be 5:00 Texas time. That's the upside of making a site that's limited to only one city.

static.config.js
Niicck: Definitely coerce it into CT. We don't want our build output to be contingent on the timezone of our build server.

static.config.js
Niicck: 🔍

src/js/helpers/cleanData.js
chiaberry: I'd still be hardcoding it though, right? `lang === 'es' ? 'date.Noon' : 'Noon'`

static.config.js
chiaberry: also thank you @robbobfrh84 for doing the heavy lifting and finding how to do it!

src/components/Pages/EventList/EventListEntry.js
```diff
@@ -0,0 +1,104 @@
...
+
+import { useMobileQuery } from 'js/helpers/reactMediaQueries.js';
+import { formatTime, formatHours } from 'js/helpers/cleanData';
+import { events as i18n } from 'js/i18n/definitions';
+
+const EventDateCalendar = ({date}) => {
+  const intl = useIntl();
+  const isMobile = useMobileQuery()
+  
+  moment.locale(intl.locale);
```
briaguya: just a note that I updated https://github.com/cityofaustin/techstack/issues/3897 to include the fact that we're using moment.js in components as well (if we're going to try to come up with a consolidated time solution at some point)

src/components/Pages/EventList/index.js
briaguya: shameless https://github.com/cityofaustin/techstack/issues/3897 plug

src/components/Pages/OfficialDocuments/OfficialDocumentPage.js
```diff
@@ -17,10 +19,9 @@ const OfficialDocumentPage = ({ officialDocuments, intl }) => {
   const maxPagesMobile = 5;
   const maxPagesDesktop = 7;
   const maxPagesShown = isMobile ? maxPagesMobile : maxPagesDesktop
-  const allDocs = officialDocuments.edges
-  const pages = buildPages()
+  const pages = buildPages(officialDocuments.edges, documentsPerPage)
   const [ pageNumber, setPageNumber ] = useState(getHash())
-  const shownPages = buildPagination()
+  const shownPages = buildPagination(pages, maxPagesShown, pageNumber)
```
briaguya: hooray reusability!

src/components/Pages/_EventList.scss
```diff
@@ -0,0 +1,239 @@
+.coa-EventListPage__Entry {
```
briaguya: overall styles question: have we abandoned colors/spacing in sass vars entirely? I know I've been magic numbering it up a bit lately and it doesn't really bother me, just wondering if we're still trying to support a sweeping "make all the blues into purples" design decision in the future or not

src/components/Pages/_EventList.scss
```diff
@@ -0,0 +1,239 @@
...
+  background-color: #0049C7;
+  color: #FFFFFF;
+}
+
+.coa-EventListPage_page.active:hover {
+  background-color: #0049C7;
+  color: #FFFFFF;
+}
+
+.coa-EventListPage_page:hover .coa-CheveronLeftBlue {
```
briaguya: definitely not something that should stop us from shipping, but it sparked my scss brain a bit and i think this could be done a little more sassily, something along the lines of:
```
.coa-EventsListPage_page {
  &:hover {
    .coa-ChevronLeftBlue {
      fill: #FFFFFF;
    }
  }
}
```

src/js/helpers/cleanData.js
briaguya: @chiaberry I don't think so, it'd be more like
```
if (time === '12:00:00') {
  return intl.formatMessage(noonKey)
}
```

src/js/helpers/cleanData.js
```diff
@@ -745,3 +747,28 @@ export const getOfferedByFromRelatedDepartments = relatedDepartments => {
 
   return offeredBy;
 };
+
+export const getEventPageUrl = (slug, date) => {
```
briaguya: just a note that url generation is happening here too

src/components/Pages/_EventList.scss
Niicck: I want to make an issue for someone to go and clean up all of our sass vars. It would be better to use them.

PR #717, opened 2020-03-09
4053 scroll to top

chiaberry:
# Description

Scroll position is being maintained even when navigating to a different page. Here's the fix:
https://reacttraining.com/react-router/web/guides/scroll-restoration/scroll-to-top

I remember there may be some weirdness with using `window` when we use it in prod. Will this work? 

https://janis-4053-scroll-position.netlify.com/

# Types of changes

# How can this be tested?

 
     

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,7 @@
-import React, { Component, Fragment, Suspense } from 'react';
-import { Root, Routes, useSiteData } from 'react-static';
+import React, { Component, Fragment, Suspense, useEffect } from 'react';
+import { Root, useSiteData } from 'react-static';
 import { Route, Switch } from 'react-router';
+import { useLocation } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -12,6 +13,17 @@ import Footer from 'components/PageSections/Footer';
 
 import 'css/coa.css';
 
+
+const ScrollToTop = () => {
+  const { pathname } = useLocation();
+
+  useEffect(() => {
+    window.scrollTo(0, 0);
```
Niicck: The bugs we'd run into in the past with "window" were with Janis builds. If you've got a root-level window in some function called by react-static, it'll freak out. Because "react-static build" is a server-side build process, and there is only a "window" object on the browser client. But since your window is safely tucked within a useEffect(), I suspect we'll be good. This is a great solution!

PR #733, opened 2020-04-11
4185 make janis preview

briaguya:
# Description
service page and official documents page previews should be working

this joplin is needed for api updates https://github.com/cityofaustin/joplin/pull/666

# How can this be tested?

 
     
* pull this branch down
* get it running (I used start-joplin-local)
* go here http://localhost:3000/en/preview/services/UGFnZVJldmlzaW9uTm9kZTox?CMS_API=https://joplin-pr-fixtures-and-preview.herokuapp.com/api/graphql and see it working

# Checklist:
- [ ] Request reviewers


src/components/PageSections/Header/index.js
```diff
@@ -10,9 +10,8 @@ import ExternalLink from 'components/ExternalLink';
 
 import LanguageSelectBar from 'components/PageSections/LanguageSelectBar';
 import HowYouKnowMenu from 'components/PageSections/HowYouKnowMenu';
-import { themePropTypes } from 'components/PageSections/Menu/proptypes';
```
briaguya, line 13: this is the first non jsprettier change

src/components/PageSections/Header/index.js
```diff
@@ -274,7 +281,6 @@ class Header extends Component {
 }
 
 Header.propTypes = {
-  navigation: PropTypes.arrayOf(themePropTypes).isRequired,
```
briaguya, line 277: this is the second non jsprettier change

src/components/_Controllers/CMSPreview.js
```diff
@@ -67,98 +48,33 @@ class CMSPreview extends Component {
     const client = createGraphQLClientsByLang(intl.locale, CMS_API);
     //TODO: one endpoint for revisions data requests
```
chiaberry, line 49: huzzah! the todo has been done!!!

src/js/queries/getPageRevisionQuery.js
```diff
@@ -0,0 +1,125 @@
...
+import guidePageFragment from './guidePageFragment';
+import informationPageFragment from './informationPageFragment';
+import locationPageFragment from './locationPageFragment';
+import officialDocumentPageFragment from './officialDocumentPageFragment';
+import topicCollectionFragment from './topicCollectionFragment';
+import topicFragment from './topicFragment';
+
+// todo in the distant future:
+// * figure out duplicate fragment issues
+// * get this working
```
chiaberry, line 14: is it not working because the fragments need to be updated?
or does it not like this format?
i want to do something similar for allPagesQuery

briaguya: an example of why it's broken:
multiple page types have contacts, we use contact fragments in multiple queries, when they're all assembled here we end up having included the contact fragment multiple times, causing errors.

Niicck: That's interesting that we're not allowed to use a fragment more than once in a query. We could make hacky fragments like I did in gql in Joplin. But hopefully there's a non-hacky solution.

PR #737, opened 2020-04-17
4289 page is part of a guide

chiaberry:
# Description

Code to inform service pages and information pages that they are part of guides. 

A question arises though around guide pages and their url. I left the original way we were formatting the guide page url in there for reference (I will remove before merging) 

it was: 
```
      // const url = '/' +
      // [
      //   guidePage.node.topics.edges[0].node.topic.topiccollections.edges[0]
      //     .node.topiccollection.theme.slug,
      //   guidePage.node.topics.edges[0].node.topic.topiccollections.edges[0]
      //     .node.topiccollection.slug,
      //   guidePage.node.topics.edges[0].node.topic.slug,
      //   guidePage.node.slug,
      // ].join('/');
```
We would just take the first topic collection and topic and just use that as the link. Guides can also be under departments (right?) and the janisInstances seem to return the department url first. If there is a preference for the topic page path instead of the department path `theme/tc/topic/guide` vs `/dept/guide`, I suppose we can get the last url in the array of instances? 

I didnt know if we did the topic url because it was easier to get than departments or if that was a conscious decision. 

# Types of changes

# How can this be tested?

 
     

# Checklist:
- [ ] Request reviewers

### DIFF
--- a/src/js/queries/getAllGuidePagesSectionsQuery.js
+++ b/src/js/queries/getAllGuidePagesSectionsQuery.js
@@ -3,30 +3,10 @@ const allGuidePagesSectionsQuery = `
     allGuidePages {
       edges {
         node {
+          id
           title
           slug
           pageType
-          topics {
-            edges {
-              node {
-                topic {
-                  slug
-                  topiccollections {
-                    edges {
-                      node {
-                        topiccollection {
-                          slug
-                          theme {
-                            slug
-                          }
-                        }
-                      }
-                    }
-                  }
-                }
-              }
-            }
-          }
           sections {
             heading
             pages {
--- a/static.config.js
+++ b/static.config.js
@@ -205,7 +205,6 @@ const getServicePageData = async (
   };
 
   if (pagesOfGuides && pagesOfGuides[servicePageData.id]) {
-    console.log(pagesOfGuides[servicePageData.id])
     // We're checking if this id is part of guide page because it may not be published and draw an error.
     servicePageData.pageIsPartOf = pagesOfGuides[servicePageData.id];
   }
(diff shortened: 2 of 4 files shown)


src/js/queries/getAllGuidePagesSectionsQuery.js
```diff
@@ -3,30 +3,10 @@ const allGuidePagesSectionsQuery = `
     allGuidePages {
       edges {
         node {
+          id
           title
           slug
           pageType
-          topics {
```
briaguya, line 9: hooray for more cleanup!

static.config.js
```diff
@@ -576,32 +577,29 @@ const getPagesOfGuidesData = async client => {
...
-            .node.topiccollection.slug,
-          guidePage.node.topics.edges[0].node.topic.slug,
-          guidePage.node.slug,
-        ].join('/');
+  allGuidePages.edges.map(async guidePage => {
+    if (guidePage.node.sections.length > 0) {
+      const guideUrl = await client.request(getPageUrlQuery, {
+        id: guidePage.node.id,
+      });
+      const url = guideUrl.allPages.edges[0].node.janisInstances[0].url;
```
Niicck, line 585: Choosing the first instance url is a good way to go. If I had to guess, I'd share your assumption that our previous url building method was chosen not out of a deliberate content strategy, but rather because it was impossible to get sensible janis url information easily.

static.config.js
```diff
@@ -203,9 +204,9 @@ const getServicePageData = async (
     offeredBy: getOfferedByFromDepartments(servicePageData.departments),
   };
 
-  if (pagesOfGuides && pagesOfGuides[id]) {
+  if (pagesOfGuides && pagesOfGuides[servicePageData.id]) {
     // We're checking if this id is part of guide page because it may not be published and draw an error.
-    service.pageIsPartOf = pagesOfGuides[id];
+    servicePageData.pageIsPartOf = pagesOfGuides[servicePageData.id];
```
Niicck, line 209: In this a new part of v3? Querying guide pages first and then checking if service/info pages are part of them? I think I remember you saying that we used to query all guide pages for every service/info pages. Or something like that.

static.config.js
briaguya, line 585: are we sure we have edges and janisInstances here? I feel like we might end up with broken builds if we don't put some checks in here

static.config.js
briaguya: this is in v2 rn https://github.com/cityofaustin/janis/blob/992901d69f3acdc4bfb66dfbc7ff01f46082ead2/static.config.js#L756

static.config.js
chiaberry: i will add some checks 👍

PR #755, opened 2020-04-30
adding condition for comma without address

robbobfrh84:
# Description

This is a bug for incomplete addresses without city, state.... 
Screen Shot 2020-04-30 at 3 31 29 PM

...And, this PR will fix it to look like 👇

Screen Shot 2020-04-30 at 5 47 21 PM

### DIFF
--- a/src/components/Contact/ContactDetails.js
+++ b/src/components/Contact/ContactDetails.js
@@ -49,7 +49,9 @@ const ContactDetailsEntry = ({
       <PhonesList phoneNumbers={phoneNumbers} />
     )}
     {email && <Email email={email} />}
-    {location && <Address location={location} />}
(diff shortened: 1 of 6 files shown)


src/components/Contact/ContactDetails.js
```diff
@@ -49,7 +49,9 @@ const ContactDetailsEntry = ({
       <PhonesList phoneNumbers={phoneNumbers} />
     )}
     {email && <Email email={email} />}
-    {location && <Address location={location} />}
+    {/* Removing incomplete location details 
+      {location && <Address location={location} />}
+    */}
```
briaguya, line 54: would this break this https://alpha.austin.gov/en/police-oversight/?
Screen Shot 2020-04-30 at 5 58 32 PM

PR #853, opened 2020-09-08
Issue 4776: make searchIndex for Elasticsearch

Niicck:
https://github.com/cityofaustin/techstack/issues/4776

# Description

### Added:
- [ ] searchIndexBuilder now builds a searchIndex that can be sent to Elasticsearch. We don't use Elasticsearch by default, and the new searchIndex is compatible with our old method of embedding directly into the Search page, like we had it before.
- [ ] The new searchIndex builder makes it a little easier to add new content types and pick the specific fields that we want to use for each content_type.
- [ ] AllPagesQuery now gets the "parentClass", "searchSummary", and "pageId" for every page. These are helpful for building the Elasticsearch searchIndex.
- [ ] searchIndex adds "body" field for OfficialDocumentPages. This sets us up for searching through that field.
- [ ] `yarn run elasticsearch` will run an elasticsearch docker image! This doesn't do anything yet, but it will.

### Fixed:
- [ ] locationPage relatedServices was still querying for JanisUrl rather than JanisUrls. I'm not sure how that was ever working.
- [ ] Changed our graphql batchSize to 25 by default, up from 10. Added $batchSize variable to change this in the future. 

Joplin PR: https://github.com/cityofaustin/joplin/pull/809

# Types of changes

# How can this be tested?
https://janis-v3-4776-elastic.netlify.app/en/search/#page=1
This search should work the same as it did before. 
- [ ] Make sure all the page results we'd expect to see show up.
- [ ] And that those links work.
- [ ] And that the right summary/data below the link shows up for correctly for each content_type.

# Checklist:
- [ ] Request reviewers


src/js/helpers/searchIndexBuilder.js
```diff
@@ -1,123 +1,287 @@
...
+          })
+        }
+      })
+      console.log(`There were errors building index ${indexName}`)
+      console.log(erroredDocuments)
+    } else {
+      console.log(`Successfully built index ${indexName}`)
+    }
+  } catch (err) {
+    console.log(`There were catastophic errors building index ${indexName}`)
```
chiaberry, line 281: catastrophe!!! really gets the point across

static.config.js
```diff
@@ -716,24 +716,30 @@ const makeAllPages = async (langCode, incrementalPageId) => {
...
   console.log(`- Building routes for ${path}...`);
 
   const client = createGraphQLClientsByLang(langCode);
 
   let pages = [];
   let after = '';
+  const batchSize = Number(process.env.REACT_STATIC_BATCH_SIZE) || 25
   while (true) {
-    const siteStructure = await client.request(allPagesQuery, { after: after });
+    const siteStructure = await client.request(allPagesQuery, { after: after, batchSize: batchSize});
```
chiaberry, line 728: i like this

static.config.js
```diff
@@ -716,24 +716,30 @@ const makeAllPages = async (langCode, incrementalPageId) => {
...
     if (!siteStructure.allPages.pageInfo.hasNextPage) {
       break;
     }
   }
 
-  // Build search index here before pages is altered.
-  const searchIndex = searchIndexBuilder(pages);
+  /**
+    Build search index here before pages are altered.
+    Hardcoding "withElasticsearch" and "indexName" for now.
```
chiaberry, line 738: hardcoded for now, how will it get toggled in the future?

Niicck: I started writing a response and then I realized that it should just be written in an issue. Then I realized it should actually be written in an epic. https://github.com/cityofaustin/techstack/issues/4919

src/js/helpers/cleanData.js
```diff
@@ -94,9 +94,9 @@ export const cleanContact = contact => {
   return cleaned;
 };
 
-export const cleanLocationPageJanisUrl = janisUrl => {
-  if (!!janisUrl.length) {
-    return janisUrl[0];
+export const cleanLocationPageUrl = janisUrls => {
+  if (!!janisUrls.length) {
+    return janisUrls[0];
```
robbobfrh84, line 99: Don't know why this worked either! But makes sense to be urls, due to the fact it's an array.

robbobfrh84 (approved):
Looks good!

src/js/helpers/cleanData.js
chiaberry: The reason it worked before is that in prod right now the schema for ServicePageNode is
```
class ServicePageNode(DjangoObjectType):
    page_type = graphene.String()
    janis_url = graphene.List(graphene.String)
    owner = graphene.Field(OwnerNode)
```
so it is janis_url on the relatedService. your joplin refactoring fixes it

Niicck: Oh interesting. Thanks for resolving that mystery for me! I couldn't figure out how it was working before.

src/js/helpers/searchIndexBuilder.js
Niicck: There are some errors that elasticsearch's API is built to handle. If there are errors outside of that, then I called them catastrophic, and the rest of the build should fail.

